PR #1, opened 2016-01-04
Update chapter.xml

Mak-Di:
Typo fix


AMDmi3:
Committed. Next time please do not change encoding of the whole file, and preferably use supported patch/problem submission services such as [bugzilla](https://bugs.freebsd.org/) or [phabricator](https://reviews.freebsd.org/).

PR #2, opened 2016-05-29
Update Handbook: default kernel support exists for rctl as of FreeBSD 10.2

yonas:
Update section 13.13.2. "Enabling and Configuring Resource Limits" of the Handbook to mention that kernel support exists for rctl as of FreeBSD 10.2.


uqs:
Please file this as a PR, see https://www.freebsd.org/support/bugreports.html

PR #3, opened 2016-05-30
Update Handbook: use ISC DHCP 4.3 instead of 4.2

yonas:
Replace "net/isc-dhcp42-server" and "net/isc-dhcp42-relay" with "net/isc-dhcp43-server" and "net/isc-dhcp43-relay".


uqs:
Please file this as a PR, see https://www.freebsd.org/support/bugreports.html

PR #5, opened 2017-04-11
Give priority to IPFW

kellytk:


brd:
What justification is there for switching these around?  Especially when this is a huge amount of churn..

kellytk:
@brd Discussion in #freebsd on Freenode found renewed interest in emphasizing use of ipfw over the alternatives.  I offered to move it to the front of the line in the handbook to convey that to new users.  The diff appears more complicated than the changes I made, which were to move the ipfw section above the pf section, and to swap the order of mentions of ipfw and pf in another place.  Do you have a suggestion for how I could do it better?

kellytk:
@brd Additional rationale provided by rgrimes: "I comes before P in the alphabit and we usually put stuff in alphabetic order. Another reason it should be first is it was first in freebsd, it has been here longer than any of the other firewall code."

uqs:
@kellytk would you be interested in reviving this after the asciidoc conversion? I think the ipf and ipfw chapters should live in separate files, then it's easier to reorder them (but note, I'm not a doc committer myself)

kellytk:
@uqs Gladly.

uqs:
@kellytk Thanks!

I'll close this PR here then, as we will soon update the hashes in this repo forcefully to the new, canonical one. I'm not sure how far along the move to asciidoc is. Once done, please re-submit your suggested edit. Thanks!

PR #6, opened 2018-01-16
Fix broken example

dlangille:


sevan:
Looks like this change has landed.

PR #39, opened 2021-03-29
Typos

cpeticus:
This now makes sense.


fel1x-developer:
According to here(https://wiki.freebsd.org/GitHub), pull requests are not officially accepted. Please summit your bug in bugs.freebsd.org. I made a patch and you can track the status here(https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=255640)

allanjude:
Slightly modified version committed as 9e73ebf001b7b6526ccf2059c8d15fdae183f762

PR #42, opened 2021-05-21
books/porters-handbook/: More reliable rebase

gogolok:
When the reader follows the git instructions, the call to git pull
--rebase will just fail with the following message most of the time:

$ git pull --rebase
error: cannot pull with rebase: You have unstaged changes.
error: additionally, your index contains uncommitted changes.
error: please commit or stash them.

In order to fix the situation in most cases, using --autosquash seems
more reliable thing to do.


bsdimp:
This change is no longer relevant. The porting guide was updated in a different way.

bsdimp:
cb632051dc2b has the other changes.

PR #46, opened 2021-09-18
Fixed a small typo in The FreeBSD Funnies

vedranmiletic:
Added space after Q::

### DIFF
--- a/documentation/content/en/books/faq/_index.adoc
+++ b/documentation/content/en/books/faq/_index.adoc
@@ -2907,7 +2907,7 @@ Just think about footsteps in wet cement.
 === How cool is FreeBSD?
 
 [qanda]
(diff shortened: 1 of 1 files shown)


dbaio:
Hi @vedranmiletic, thanks for the patch.

16.1 and 16.2 are not using the proper way of question and answers from Asciidoctor.

Please, as you are here, change it to the proper format:
```
[qanda]
What is the answer?::
This is the answer.

Have you seen my duck?:: No.
```
https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/

dbaio:
Asciidoctor Question and Answer Lists in this context don't look good.

I've changed it to a regular paragraph.

Now it's rendering as the [legacy format](https://www-legacy.freebsd.org/doc/en/books/faq/funnies.html).

And thanks for your report.

vedranmiletic:
Apologies for the delay in responding, I must have missed the notification. Thanks for fixing the issue.

PR #50, opened 2022-01-10
Fixed compile issue

blitztide:
Fixed issue when compiling where module.h has to be delared after param.h

BUGZILLA ref: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=261014


dbaio:
Hi.

Committed in d00b3eb7681921e974ca687151eea5f42fe81b53.

Thanks for your contribution.

PR #54, opened 2022-01-28
added BSD abbreviation information.

D4n2021:


debdrup:
I don't think this belongs in the FAQ - if anything, it belongs in the introduction of the handbook.

Moreover, there's some confusion over whether it's Berkeley Software Distribution, Berkeley Standard Distribution, or Berkeley System Distribution; part of this, I'm sure, is due to 1BSD and 2BSD starting out as mere collections of software (ie. the vi editor and a Pascal compiler, and some other bits of software) that was meant to be used with Research UNIX, whereas 3BSD added a complete kernel, many more userland components, and 4BSD et al added even more things.

pauamma:
> I don't think this belongs in the FAQ - if anything, it belongs in the introduction of the handbook.

Agreed. And even if worth including in the FAQ, I don't think a question+answer that's only tangentially related is the right place.

dbaio:
Agreed with @debdrup and @pauamma.

Please reopen this or open a new PR if you have any other thoughts.

Or send an email to the [FreeBSD Doc mailing list](https://lists.freebsd.org/subscription/freebsd-doc).

PR #61, opened 2022-04-08
Update _index.adoc

fearedbliss:


bsdimp:
We should apply this. <approved> for what it's worth

fearedbliss:
@bsdimp You mean like accept the merge request? What's the recommended way for me to submit patches in the future? I looked at the other closed pull requests and it seemed like there was activity to submitting changes via github haha.

bsdimp:
I'm mostly tagging for others... I tried to cherry-pick this to do the merge by hand and I got an error which is odd and today is crazy enough I don't have the energy to root cause it :) Submitting here is fine.

fearedbliss:
@bsdimp Ah ok sounds good :). Yea I saw the typo and I was like .. should I even open a PR for this given that it's an "a"? But then I was like I mean if I don't do it then it may take a while for someone to find that so might as well just report it at least and see what happens. Plus the FreeBSD documentation is amazing so I definitely want to help keep it polished as much as possible.

dbaio:
Committed on 1c26b3934ac8e09f94f9b797443a258dd26ce2b9. Thanks for your contribution.

PR #62, opened 2022-04-08
Update _index.adoc

fearedbliss:


fearedbliss:
I hope nobody minds the multiple independent submissions to these typos. I believe this align with regular coding practices of having separate commits for separate pieces of code in order to facilitate tracking and also easier reverts.

I also noticed this issue in my browser (Brave), there seems to be some floating issues in the freebsd markdown rendering engine perhaps? We can move this to the appropriate place if anything. Let me know!

`Brave Version 1.36.122 Chromium: 99.0.4844.88 (Official Build) (64-bit)`

image

The github preview (still in Brave) shows that section correctly however. So the markdown itself seems to be correct. This is a separate issue of course but in the same block this PR is modifying stuff in.

image

This is the same section in `Firefox 99`:

image

dbaio:
Committed on 51103352a58fccc97f1258e14b0d6a27832a21a9. Thanks for your contribution.

PR #66, opened 2022-04-15
Info for serving poudriere packages privately

fearedbliss:
This adds a small section to the "Configuring pkg Clients to Use a Poudriere Repository" showing the user how to point to their packages repository (package directory in this case) via `file://`. Allowing the user to not have to publicly expose their packages over the internet.

### DIFF
--- a/documentation/content/en/books/handbook/ports/_index.adoc
+++ b/documentation/content/en/books/handbook/ports/_index.adoc
@@ -1301,6 +1301,16 @@ custom: {
 }
 ....
 
+However, if you don't want to expose your packages publicly, and you have access to the poudriere packages directory (either because you are using poudriere on the same system that built these packages, or you've exposed it over the network with NFS), you can make your package repository point directly to your package directory:
+
+[.programlisting]
+....
+custom: {
+	url: "file:///usr/local/poudriere/data/packages/11amd64",
+	enabled: yes,
+}
+....
+
 [[ports-nextsteps]]
 == Post-Installation Considerations
 


pauamma:
LGTM

debdrup:
In order to credit you properly with `git commit --author "Jonathan Vasquez <jon@xyinn.org>"`?

fearedbliss:
@debdrup Thank you. That's correct :).

debdrup (changes requested):
Alright, I've had a little time to review it now, and I've found a couple of things.

Firstly, [we have some writing style guidelines](https://docs.freebsd.org/en/books/fdp-primer/book/#writing-style) that I think need to be applied to this.

Secondly, I think it needs to be broken into multiple sentences and have the parenthesis moved - it's a bit too run-on as it is, and the parenthesis feels more like a conditional clause that could be worded better.

Overall, I think it's a good change though - so if you can run over the FDP (not just the section I linked to), adjust as necessary, and keep in mind the other thing, I think we can get this landed in short order. :)

fearedbliss:
@debdrup Thanks for the feedback :). I'll take a look at the guidelines and push another change.

fearedbliss:
@debdrup I finally got a chance to take a look at the FreeBSD guidelines and philosophy for writing. After looking at it, I've realized that my personal philosophy for technical writing is very different (and I highly prefer writing informally since I like to feel (and want the reader to feel) closer). There's many other things as well but that's one of the main ones. For this reason I've decided to close this PR. If someone really wants to include this change then I encourage them to take what I wrote and convert it to the FreeBSD style appropriately.

Thanks for taking the time to take a look at my PR and give me feedback. I felt it was necessary to at least give a reason as to why I would continue rather than just silently close it.

Take care!

fearedbliss:
@debdrup Hey, since recently I got a chance to spend a lot more time in the docs (mostly reading), I re-read most of the FDP again and decided to take another crack at this. I checked out a copy of the docs locally and decided to open the following PR:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=263955

We can continue further discussion on there since this branch is stale anyways and I generated a diff for submission via the problem report interface instead.

I'll be using this experience as an opportunity to further improve my "formal technical writing" as oppose to my "informal technical writing" :D.

PR #74, opened 2022-05-19
Remove duplicated AMD64 Virtual Machine Images link

hartraft:
AMD64  Virtual Machine Images link for FreeBSD 13.1-RELEASE is referenced twice, the link to the CI-IMAGES doesn't appear in the previous 13.0-RELEASE links so I removed that one.


dbaio:
Hi.

I've talked with the re team, and this content is valid; the issue is that it's in a different directory structure, so the link name will be renamed and not removed.

Anyway, thank you for taking the time to help with this.

PR #90, opened 2022-10-25
[32.3.4.1.1. How to Find Access Points] Update _index.adoc

mezantrop:
As noted below: "The interface must be `up` before it can scan", so under certain **proven** conditions, "up" and "scan" do not work together and should be run separately.

### DIFF
--- a/documentation/content/en/books/handbook/advanced-networking/_index.adoc
+++ b/documentation/content/en/books/handbook/advanced-networking/_index.adoc
@@ -530,7 +530,8 @@ Only the superuser can initiate a scan:
 [source,shell]
 ....
 # ifconfig wlan0 create wlandev ath0
-# ifconfig wlan0 up scan
+# ifconfig wlan0 up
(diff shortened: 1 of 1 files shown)


mezantrop:
No commentaries, reactions, objections?

sergio-carlavilla:
Fixed in https://cgit.freebsd.org/doc/commit/?id=3c6d3dea4a3ee60e7f0033afc9c5bf74e9ae1d31
Sorry for the delay, if this happens again, please ping me in "carlavilla@FreeBSD.org"

mezantrop:
> Fixed in https://cgit.freebsd.org/doc/commit/?id=3c6d3dea4a3ee60e7f0033afc9c5bf74e9ae1d31 Sorry for the delay, if this happens again, please ping me in "[carlavilla@FreeBSD.org](mailto:carlavilla@FreeBSD.org)"

Thank you very much :)

PR #93, opened 2022-11-07
Remove invalid ports and links in Handbook

ykla:
Just remove invalid ports and links in Handbook. The links or ports that can be replaced I have already done so.


lwhsu:
https://reviews.freebsd.org/D37322

PR #92, opened 2022-10-31
Update ZFS arc min/max sysctl loader variables

xor-gate:
Update arc min and max loader syctl variable names based on 13.1-RELEASE-p2

PR #98, opened 2022-11-28
Fix typos in ZFS arc legacy sysctl tunables

xor-gate:
This is a followup of https://github.com/freebsd/freebsd-doc/pull/92

Seems there where some leftover underscores for italic font which where not probably formatted.


dbaio:
Committed on 6148547a289d0a901b832518ef7cb304b36c8ba1.
Thanks for your contribution.

PR #100, opened 2022-12-11
         pr-guidelines: begin corrections and updates

grahamperrin:
Context: 'Bug Reports' <https://www.freebsd.org/support/bugreports/> describes 'Problem Report Handling Guidelines' <https://docs.freebsd.org/en/articles/pr-guidelines/> as 'useful background information'. 

Status, not state. 

Deal with some non-existent statuses. Much more rewriting is required in areas such as this. 

Minimise use of the PR acronym. 

AsciiDoc line breaks, et cetera.

### DIFF
--- a/documentation/content/en/articles/pr-guidelines/_index.adoc
+++ b/documentation/content/en/articles/pr-guidelines/_index.adoc
@@ -5,7 +5,7 @@ authors:
   - author: Hiten Pandya
 description: These guidelines describe recommended handling practices for FreeBSD Problem Reports (PRs).
 trademarks: ["freebsd", "general"]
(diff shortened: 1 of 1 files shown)


grahamperrin:
[⚙ D37671 Problem Report Handling Guidelines (pr-guidelines): correct and update, or archive](https://reviews.freebsd.org/D37671)

grahamperrin:
From D37671 in Phabricator: 

> … For FreeBSD Bugzilla as is: so much of the article is wrong, I can not describe it as useful.

> … archive the article, cease mentioning it at pages such as Bug Reports, and abandon this revision. …

I don't doubt that the article was largely valuable in an era when it would have been correct (at its essence), however that era has passed. 

For various reasons, I strongly recommend archiving. 

Closed (abandoned) in Phabricator. Closing here.

PR #105, opened 2022-12-18
Events, 2022: fix FreeBSD Fridays and FreeBSD Office Hours

grahamperrin:
After discovering that the most recent event began at 18:00 (not 17:00, as advertised at the FreeBSD home page), I took a closer look at the `.toml` file entries for the two series.


grahamperrin:
[⚙ D37735 Events, 2022: fix FreeBSD Fridays and FreeBSD Office Hours](https://reviews.freebsd.org/D37735)

PR #106, opened 2023-01-08
Corrected broken links to /support/bugreports/

vedranmiletic:
English locale used links prefixed with `/en/`, which are broken. This removes the prefix, which is correct.

Japanese locale also used links prefixed with `/en/`. This changes the prefix to `/ja/`, which is correct.


vedranmiletic:
Sorry, messed up branches in my repo, will re-submit.

PR #107, opened 2023-01-09
Add tip: ftp.freebsd.org automatic routing

gaelanlloyd:
I found out that selecting the `ftp.freebsd.org (Main site)` item will automatically route you to the nearest/fastest mirror. I believe this would be a helpful item to have documented.


dbaio:
That's true.

Note that all `ftp.freebsd.org` references will soon point to `download.freebsd.org`.
And we will need to update this chapter/screenshots/etc.

It was already updated on CURRENT/main (https://github.com/freebsd/freebsd-src/commit/591e29117964248d4bfea639e973a12e27f96ac9) but not in FreeBSD 12/13.

So for me, it looks good at this moment.

And thanks for working on this.

dbaio (approved):

dbaio:
Committed on https://github.com/freebsd/freebsd-doc/commit/ff91bafb0ac1decb30b0694488d4cc845f547cd5. Thank you.

PR #108, opened 2023-01-18
Cleaned up extra whitespace in HTML in TOML files

vedranmiletic:


sergio-carlavilla:
Fixed in https://cgit.freebsd.org/doc/commit/?id=17ca76d847606c99ffebad98b9e5636a26b6524b
Sorry for the delay.

PR #109, opened 2023-01-18
Corrected broken links to /support/bugreports/

vedranmiletic:
English locale used links prefixed with `/en/`, which are broken. This patch removes the prefix, which is correct.

Japanese locale also used links prefixed with `/en/`. This changes the prefix to `/ja/`, which is correct.


sergio-carlavilla:
Fixed in https://cgit.freebsd.org/doc/commit/?id=6cae6054f814acd672a4f443a4b8d727b4ee4863
Thanks!

PR #110, opened 2023-01-21
Improve documentation build instructions

gaelanlloyd:
This new content addresses knowledge gaps that weren't immediately obvious to me as a first-time FreeBSD documentation contributor.


sergio-carlavilla:
Sorry for the delay and thanks for your contribution.
Just made, with a few changes in: https://cgit.freebsd.org/doc/commit/?id=1aadaf5e

gaelanlloyd:
Thanks! It looks like there’s a slight formatting issue here with a mismatched backtick:

DD4A2FE7-7023-4E11-9017-D4B40B990C6B

sergio-carlavilla:
Fixed! Thanks!

PR #111, opened 2023-02-15
Vertically align table contents to improve position of step numbers

gaelanlloyd:
Top-aligned table cell contents would make [sections with step numbers](https://docs.freebsd.org/en/books/handbook/jails/#jails-install-source) easier to scan through. The numbers would line up with the beginning of the step text.

I can't figure out how to recompile the Sass to test this out, and I'm sure making a change of this scale will likely impact other tables, so I'd appreciate review and feedback on this change before it's implemented.

Current:

Proposed:


bsdimp:
I like the results. No clue if this is the right way to get them.

PR #114, opened 2023-02-19
Nvidia fixes

shkhln:
1. 304, 340, 390 are out of support: https://nvidia.custhelp.com/app/answers/detail/a_id/3142/~/support-timeframes-for-unix-legacy-gpu-releases. In addition to that 304 and 340 are also completely broken.
2. The nvidia-modeset advice is simply _wrong_.

### DIFF
--- a/documentation/content/en/books/handbook/x11/_index.adoc
+++ b/documentation/content/en/books/handbook/x11/_index.adoc
@@ -259,12 +259,6 @@ Those with older cards will have to check below which version supports them.
 |===
 | Package | Supported hardware
 
-| x11/nvidia-driver-304
-| link:https://www.nvidia.com/Download/driverResults.aspx/123712/en-us/[supported hardware]
-
-| x11/nvidia-driver-340
-| link:https://www.nvidia.com/Download/driverResults.aspx/156167/en-us/[supported hardware]
-
 | x11/nvidia-driver-390
 | link:https://www.nvidia.com/Download/driverResults.aspx/191122/en-us/[supported hardware]
 
@@ -275,11 +269,6 @@ Those with older cards will have to check below which version supports them.
 | link:https://www.nvidia.com/Download/driverResults.aspx/187164/en-us/[supported hardware]
 
 |===
-
-[WARNING]
-====
-Version 304 of the NVIDIA(R) graphics driver (nvidia-driver-304) does not support xorg-server 1.20 or later.
-====
 
 The latest NVIDIA(R) driver can be installed by running the following command:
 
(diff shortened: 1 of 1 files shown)


grahamperrin:
Related: 

* [269626 – FreeBSD Handbook: NVIDIA: corrections](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=269626)
* <https://wiki.freebsd.org/Graphics#NVIDIA_graphics>

I'll update the wiki in due course.

shkhln:
Is there a reasonable time frame for the "due course"? As you are well aware, I publicly promised to drop my ports if this is not merged and we need some date for this.

shkhln:
> drop my ports

Just in case somebody thinks this is too harsh: we are already ignoring the intention behind for the nvidia-driver pkg message, actual people experience and even the confirmation from one of the Nvidia Unix driver devs.

sergio-carlavilla:
I asked about this to the current nvidia maintainer in FreeBSD.
I'll make the commit when I have an answer from him.
Thanks for the patch and sorry for the delay

grahamperrin:
> Is there a reasonable time frame for the "due course"?

There's the interim change that was made around two minutes after <https://github.com/freebsd/freebsd-doc/pull/114#issuecomment-1436524555>. I intended to await a commit here before making a more substantial change.

danfe:
> In addition to that 304 and 340 are also completely broken.

That's somewhat an overstatement.  While nVidia had apparently [neglected](https://forums.developer.nvidia.com/t/please-recompile-304-for-xorg-server-1-20-and-freebsd-i386/119963) an earlier request to catch up the -304 branch with xserver ABI 24, there are [fresh reports of success](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=266682#c3) with -340 version by ignoring the ABI:
```
Section "ServerFlags"
Option "IgnoreABI" "1"
EndSection
```
> The nvidia-modeset advice is simply _wrong_.

This is true.  The `nvidia-modeset` kernel module was introduced in driver version 358.09 to program the display engine of the GPU (changing resolutions, transformation settings, display device hot-plugging, etc.) and works in conjunction with the `nvidia.ko` kernel module.  At the time of the writing, only the latest version of the driver shipped this module, most users would still go the `kld_list="nvidia"` way.  We simply forgot to flip the switch as the time had passed.  I'll update the port's `pkg-message` accordingly.

shkhln:
> > we are already ignoring the [intention](https://reviews.freebsd.org/D7700) behind the nvidia-driver pkg message
> 
> We're not ignoring anything. If you read it carefully, you'd notice that commit you reference largely quotes the `pkg-message` text.

Not you, those two guys. I didn't want to make this personal, thus "we".

> 
> > The nvidia-modeset advice is simply _wrong_.
> 
> This is true when it is advised _out of the context_, I presume that's the problem with the Handbook?

No, the problem with the Handbook is that now it downright tells users to _avoid_ nvidia-modeset. There is no nuance there at all.

sergio-carlavilla:
I'll upgrade the handbook information with the commit https://cgit.freebsd.org/ports/commit/?id=a09185f8b66c
Thanks @danfe :)

sergio-carlavilla:
Fixed in https://cgit.freebsd.org/doc/commit/?id=9af61238

grahamperrin:
Likewise, thanks :-)

> I'll update the wiki in due course.

Done. 

For updates to the FreeBSD Handbook, I might put something in Phabricator this weekend.

PR #118, opened 2023-03-03
Fix Handbook OpenSSH Examples

tomhukins:
The Handbook contains some obsolete and unnecessary uses of SSH.  These two commits fix them.

### DIFF
--- a/documentation/content/en/books/handbook/security/_index.adoc
+++ b/documentation/content/en/books/handbook/security/_index.adoc
@@ -1747,14 +1747,11 @@ The following command tells `ssh` to create a tunnel for telnet:
 
 [source,shell]
 ....
-% ssh -2 -N -f -L 5023:localhost:23 user@foo.example.com
+% ssh -N -f -L 5023:localhost:23 user@foo.example.com
 %
 ....
 
 This example uses the following options:
-
-`-2`::
-Forces `ssh` to use version 2 to connect to the server.
 
 `-N`::
 Indicates no command, or tunnel only.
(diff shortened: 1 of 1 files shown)


documentation/content/en/books/handbook/security/_index.adoc
```diff
@@ -1913,7 +1910,7 @@ To force the users to use keys only, the following option may be configured:
 
 [.programlisting]
 ....
-AuthenticationMethods publickey
```
sergio-carlavilla, line 1916: I'm gonna commit all except this, based on the information we have just before is correct, in this way we force users to make the login with the public key

tomhukins: Thank you for your helpful review.  You're right that the Handbook in its current form is correct: I'm unsure why I believed otherwise.  I've pushed an updated version of my branch that contains only the one desirable commit.

sergio-carlavilla:
Thanks for the patch, about the author, are you ok with "Tom Hukins <hello@exceptone.co.uk>"

tomhukins:
> Thanks for the patch, about the author, are you ok with "Tom Hukins [hello@exceptone.co.uk](mailto:hello@exceptone.co.uk)"?

No, please don't use that address: I never use it for committing.

I prefer to use my FreeBSD address, tom@FreeBSD.org, as included in the commit I submitted, but if you won't use that (I can't think why - the freebsd-src committers have no problem using it), use tom@eborcom.com instead.  Thank you.

sergio-carlavilla:
Fixed in https://cgit.FreeBSD.org/doc/commit/?id=bb022d98

Thanks!

PR #124, opened 2023-03-12
Status report process: template: simplify access

grahamperrin:
Move the template from the static status directory, to the main status directory.

Edit the process document, to allow online views of the template (no need to download).


grahamperrin:
Preview: 

* <https://github.com/grahamperrin/freebsd-doc/blob/status/documentation/content/en/articles/freebsd-status-report-process/_index.adoc#16-submit-your-report>.

lsalvadore (commented):
Thanks @grahamperrin, but I disagree with this change for the following reasons:

- it creates a rendered webpage at https://www.FreeBSD.org/status/report-sample/, which we do not need;

- it actually makes access worse. The template is meant to be downloaded and modified. The links you suggest need an extra click to get to the raw downloadable file. You could improve acces by replacing them with the link to the raw downloadable file however. But why should we go through Phabricator or GitHub to get the file? The first platform is likely to change someday (also because Phabricator is not actively maintained anymore), while the second is a third-party platform, which we are happy to use but I believe it is better if we are not dependent on it. We could link to the self-hosted link https://cgit.freebsd.org/doc/plain/website/content/en/status/report-sample.adoc instead, but we would still get the extra useless rendered page.

I think the solution we have implemented now is good. Maybe @debdrup, @pauamma, or @sergio-carlavilla also want to express their opinion on this.

grahamperrin:
Ah, 

> * The template is meant to be downloaded and modified.

The essence of this PR was to avoid unnecessary download. (It's not necessary for modification.)

lsalvadore:
We would need to know what are the workflows of our report authors to see if they prefer to download templates or to edit them online. Your suggestion assumes that authors have either a Phabricator account or a GitHub account, which is likely but not certain. And for sure, we receive reports by email all quarters (we have already received at least one this quarter too), hence there are authors who prefer to avoid GitHub (and at least one who also prefers to avoid Phabricator).

The download option seems preferable to me. It does not depend on any account. It does not even depend on a full featured web browser: you could even download it with curl.

By the way, I have seen how it is possible to edit files through GitHub, but I failed to find the option on Phabricator. I do not plan to use it (I clone the git repositories on my machine and edit them in there, then submit Phabricator reviews, GitHub pull requests or push directly), but if you could tell me where the feature is it might still be useful some day. Thanks.

grahamperrin:
Thanks, I think we can safely close this PR. It's possible/likely that I have a peculiar way of working with/around GitHub!

@lsalvadore as far as I know, you're correct about editing with Phabricator. (There's the _suggested edit_ feature when reviewing, but it's not like editing an entire file.)

I do think that the template file could/should be moved to the main `status` directory, but that alone isn't a reason to leave this open.

PR #128, opened 2023-03-16
Update _index.adoc

jdpc86:
per https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270229, removing invalid link for now until a replacement is in place


sergio-carlavilla:
Fixed in: https://cgit.freebsd.org/doc/commit/?id=d84a170606cd2b62ef11c22a5848c4a81956bfca
Thanks!

PR #129, opened 2023-03-16
Update _index.adoc

jdpc86:
per https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270229, removed invalid link for now until a replacement becomes available


sergio-carlavilla:
Fixed in: https://cgit.freebsd.org/doc/commit/?id=1d52bc22a75657661e60adfa531825dac76ee93b
Thanks!

PR #133, opened 2023-03-17
developers-handbook: ipv6: Use ping -6 instead of ping6

jlduran:
Although the command `ping6` is still supported, its man page has been disconnected from the build.

Replace occurrences of `ping6` with `ping -6`.


sergio-carlavilla:
Fixed on https://cgit.freebsd.org/doc/commit/?id=86d39751
Thanks!

PR #152, opened 2023-04-01
handbook: Add GitHub as another source for src pull requests

jlduran:
As mentioned in CONTRIBUTING.md, GitHub is also another source for src pull requests.


bsdimp:
Seems good to me.

PR #161, opened 2023-04-08
status: 2023q1: Foundation: last, this …

grahamperrin:
Consider the last report, entitled _FreeBSD Status Report Fourth Quarter 2022_. 

Expect this overall report to fall under a main heading (and HTML page title): 

> # FreeBSD Status Report First Quarter 2023

In this context – this report for this (first) quarter – the _last_ quarter is 2022q4 (that's what I thought when reviewing <https://reviews.freebsd.org/D39461>). 

jrm @Jehops clarified at <https://reviews.freebsd.org/D39461#898412>.  

Reduce ambiguity here. 

_During the first quarter of 2023, …_ is already non-ambiguous. 

Where _Last quarter_ is changed to _This quarter_, it's clear enough because it's in the midst of paragraphs that are solely about 2023.

### DIFF
--- a/website/content/en/status/report-2023-01-2023-03/freebsd-foundation.adoc
+++ b/website/content/en/status/report-2023-01-2023-03/freebsd-foundation.adoc
@@ -30,7 +30,7 @@ When we received our first million dollar donation, the plan was to use up to 10
 
 The 2023 budget is in the process of being approved by the board of directors and will be published once it is approved.
 
-Last quarter we received donations from Juniper, Tarsnap, Microsoft, and Stormshield.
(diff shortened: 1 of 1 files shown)


lsalvadore:
I will let @Jehops decide what is better here. To solve ambiguities about quarters, the best solution is maybe to simply tell what the quarter is: instead of using _last_ and _this_, we can use _2022q4_ and _2023q1_.

Jehops:
@grahamperrin Those changes took good to me.

grahamperrin:
Thank you both.

PR #166, opened 2023-04-11
13.2R/announce/: minor corrections, line breaks

grahamperrin:
Fix/change four things:

> … months from the 13.2 release date, which is yet to be scheduled at
> the time of this writing.

> … downloaded via https from …

> … root, which it is strongly recommended to change the password for
> both …

– and uppercase for ARM.

Whilst here: one sentence per line.

Reviewed by:           xxx
Approved by:           xxx
Fixes:                 xxx
Differential revision: xxx

### DIFF
--- a/website/content/en/releases/13.2R/announce.adoc
+++ b/website/content/en/releases/13.2R/announce.adoc
@@ -17,7 +17,8 @@ include::shared/releases.adoc[]
 
 Date: {thisDate}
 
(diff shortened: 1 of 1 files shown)


grahamperrin:
@cperciva I'll be at my day job for a few hours, would you like this PR to be echoed to Phabricator? 

Only one noticeable mistake, under <https://www.freebsd.org/releases/13.2R/announce/#_support>: 

> … the 13.2 release date, which is yet to be scheduled at the time of this writing.

The other things are less noticeable, but may as well receive attention.

PR #168, opened 2023-04-13
releases: Fix tzdata version

jlduran:
Commit https://github.com/freebsd/freebsd-src/commit/57472c9a30abce230797fcf1e9435fb603d7425e updated to tzdata 2023c.


sergio-carlavilla:
Fixed in: https://cgit.freebsd.org/doc/commit/?id=32a53605f4e7f66ef83ccf6ba3b874cff4d44808

I added the erratas to the errata.adoc file

PR #180, opened 2023-04-29
Korean translation from Now

cygnus6960:
Most of the handbook, and small part of the articles and developers-handbook. 

This translation was done with the most recent update.

@dbaio


dbaio:
Hello.

Thanks for working on translations.
A few points here before having your work committed. 

Please, submit a PR for each document.

The commits and PRs would be like this (as an example):

- ko/articles/bsdl-gpl: Sync with en
or
- ko/articles/bsdl-gpl: Update translation


and

- ko/books/handbook: Add partial translation


The developers handbook seems incomplete, I would suggest to focus first on articles/bsdl-gpl and books/handbook.

For the books/handbook, you need to include all .po and .adoc files; otherwise, the build will fail.
The parts not translated yet will remain in English.

If you have any questions, please let me know.

Regards.

dbaio:
I've pushed only the ko/articles/bsdl-gpl part on 84a3c57be850f3760443f846a7328dae4785b624.

Please, open another one for the Handbook when you are ready.

Thank you!

PR #181, opened 2023-05-04
shared/en/mailing-lists.adoc: improvements

grahamperrin:
Consistency, and more.

The need for this became apparent whilst drafting https://github.com/freebsd/freebsd-doc/commit/e63798df73d7d956d8c3ce3f81225421fdd1db23 … there remains a mismatch between: 

* 'the FreeBSD ACPI mailing list' at <https://docs.freebsd.org/en/books/handbook/config/#ACPI-submitdebug> and elsewhere
* the broader, true name 'ACPI and power management development' at <https://lists.freebsd.org/archives/freebsd-acpi/>.

### DIFF
--- a/shared/en/mailing-lists.adoc
+++ b/shared/en/mailing-lists.adoc
@@ -7,11 +7,11 @@
 :mailing-lists-url: https://lists.freebsd.org
 :mailing-lists: {mailing-lists-url}[{mailing-lists-desc}]
 
-:freebsd-accessibility-desc: FreeBSD accessibility discussions
+:freebsd-accessibility-desc: FreeBSD accessibility mailing list
 :freebsd-accessibility-url: https://lists.FreeBSD.org/subscription/freebsd-accessibility
 :freebsd-accessibility: {freebsd-accessibility-url}[{freebsd-accessibility-desc}]
 
-:freebsd-acpi-desc: FreeBSD ACPI mailing list
+:freebsd-acpi-desc: FreeBSD ACPI and power management development mailing list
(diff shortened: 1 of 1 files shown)


lwhsu:
IMHO we could just drop all the "mailing list(s)" in the description.

lwhsu:
Also we need to keep the consistency with the description with https://lists.freebsd.org , if possible, a source of truth is better.

grahamperrin:
Thanks, 

> IMHO we could just drop all the "mailing list(s)" in the description.

I had similar thoughts. We'd need a sweep of related texts for grammar, e.g. to remove the 'the' prefix. 

----

There's also the possibility/notion of _expanding_ some descriptions to begin with 'the', for example: 

https://github.com/freebsd/freebsd-doc/blob/ba1567988437ff28a92785b4dd49aa627841cc33/shared/en/mailing-lists.adoc#L10

> FreeBSD accessibility discussions

– could become: 

> the FreeBSD accessibility mailing list

----

For now, I'm deliberately not giving too much thought to these things … ultimately I'd like an end result to be **memorable** – without the need for writers to call up the shared file whenever they need a reminder. 

IMHO the current approach (not this draft of the PR) is not memorable. YMMV

grahamperrin:
> … We'd need a sweep of related texts for grammar, e.g. to remove the 'the' prefix. …

For now, I take an ad-hoc approach to fixing omissions of 'the' when I stumble across them. Yesterday's example: 

- https://github.com/freebsd/freebsd-doc/commit/0940974b0a430198f4306ffbaec0af9aeac006b2

PR #183, opened 2023-05-09
articles/hubs: looking for official mirrors

khorben:
Hi there, I have been told the FreeBSD project is now looking for official mirrors (as opposed to community mirrors).
I figured I would update the documentation to that effect. About the exact procedure, ISTM that e-mailing `clusteradm@` is the way to go.

### DIFF
--- a/documentation/content/en/articles/hubs/_index.adoc
+++ b/documentation/content/en/articles/hubs/_index.adoc
@@ -56,7 +56,7 @@ toc::[]
 
 [NOTE]
 ====
-We are not accepting new mirrors at this time.
+We are not accepting new community mirrors at this time.
 ====
 
 [[mirror-contact]]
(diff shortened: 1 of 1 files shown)


documentation/content/en/articles/hubs/_index.adoc
```diff
@@ -402,7 +402,7 @@ If Mirror1 does not update for a while, lower tier mirrors will begin to mirror
 [[mirror-official-become]]
 === How to Become Official Then?
 
-We are not accepting any new mirrors at this time.
+Please contact the Source Repository Manager mailto:clusteradm@FreeBSD.org[clusteradm@FreeBSD.org].
```
lwhsu, line 405: "Source Repository Manager" should be "Cluster Administrators", and maybe a link to "/administration/#t-clusteradm" would be better since it's intended to not use full email.

khorben: Thanks, I have updated the branch to reflect this change.

sergio-carlavilla:
Fixed in: https://cgit.freebsd.org/doc/commit/?id=582049d58cf358c7514a0c9b07dc8a0a7e7f108f
Thanks!

PR #185, opened 2023-05-11
A small note about rc.conf and kld_list

mrcrilly:
I believe this small note could be beneficial. It introduces the reader to the potentially new idea of man pages, and it helps them appreciate that kld_list has a particular format and can actually be used to load multiple kernel modules.

See the #community channel on Discord @ 2030 AEST for an example of a potential failure on our part to explain this correctly (at a time when it could be important.)

### DIFF
--- a/documentation/content/en/books/handbook/x11/_index.adoc
+++ b/documentation/content/en/books/handbook/x11/_index.adoc
@@ -218,6 +218,11 @@ To install the package execute the following command:
 ....
 ====
 
+[TIP]
+====
+If you're unfamiliar with the `/etc/rc.conf` file and the `kld_list` option, then checkout the link:https://man.freebsd.org/cgi/man.cgi?rc.conf[`rc.conf`] manpage and search for `kld_list`. Given the importance of this flag, it's important you make yourself familiar with its format in case you need to load multiple kernel modules in the future.
+====
+
 [[x-configuration-amd]]
 === AMD(R)
 


documentation/content/en/books/handbook/x11/_index.adoc
```diff
@@ -218,6 +218,11 @@ To install the package execute the following command:
 ....
 ====
 
+[TIP]
+====
+If you're unfamiliar with the `/etc/rc.conf` file and the `kld_list` option, then checkout the link:https://man.freebsd.org/cgi/man.cgi?rc.conf[`rc.conf`] manpage and search for `kld_list`. Given the importance of this flag, it's important you make yourself familiar with its format in case you need to load multiple kernel modules in the future.
```
grahamperrin, line 223:
```suggestion
If you're unfamiliar with the [.filename]#/etc/rc.conf# file and the `kld_list` variable, then see the man:rc.conf[5] manual page and search for _kld_list_.

If ever you're guided to _add_ or _create_ a `kld_list` line: take care to not create two or more lines. If multiple kernel modules are required, you must list them in a single line (as explained in the manual page). 
```

grahamperrin (changes requested):
I guess, this tip arises partly from cases where people mistakenly create one `kld_list` line per module to load … true?

documentation/content/en/books/handbook/x11/_index.adoc
bsdimp:
```
The `/etc/rc.conf` file contents and its `kld_list` option are covered in the link:https://man.freebsd.org/cgi/man.cgi?rc.conf[`rc.conf`] manpage. Search `kld_list` and note this option is specified only one, but contains a list of modules to load. Directions to add or delete this option elsewhere are a shorthand for _adding_ or _removing_ modules from this list, or to _add_ it not yet present. Note that multiple lines with the same option cause the later lines to override the earlier ones, not for them to add to the list.
```

PR #188, opened 2023-05-23
Committer's Guide: four tiers: heading levels

grahamperrin:
Level 3 section title for tiers 1, 2, and 3, and for the unsupported tier.

### DIFF
--- a/documentation/content/en/articles/committers-guide/_index.adoc
+++ b/documentation/content/en/articles/committers-guide/_index.adoc
@@ -3276,7 +3276,7 @@ The native "ABI" generally shares certain properties with the kernel ABI such as
 
 Tiers are defined for both kernels and userland ABIs. In the common case, a platform's kernel and FreeBSD ABIs are assigned to the same tier.
 
-=== Tier 1: Fully-Supported Architectures
+==== Tier 1: Fully-Supported Architectures
 
 Tier 1 platforms are the most mature FreeBSD platforms.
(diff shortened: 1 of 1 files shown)


bsdimp:
I think this is good, but I'm not a doc person.

grahamperrin:
To help readers visualise the proposed change, here's the level 2 section heading that precedes the change: 

* [before](https://docs.freebsd.org/en/articles/committers-guide/#_platform_targets) (live)
* [after](https://github.com/grahamperrin/freebsd-doc/blob/patch-9/documentation/content/en/articles/committers-guide/_index.adoc#212-platform-targets) (preview in GitHub). 

Strictly speaking, the L3 headings for the four tiers should be preceded by a L2 heading that includes (or comprises) the word _tiers_. I can easily move and change some texts, a little, to improve the order and sense of things, but first: 

* let's consider the four headings alone. 

----

I began this PR because [the intent describes **four** tiers](https://docs.freebsd.org/en/articles/committers-guide/#_statement_of_general_intent) and looking at headings, I counted only **three** – three that are _numbered_. 

Essentially: 

* is _unsupported_ a tier, or not a tier? 

I can see that it was previously a numbered tier, but now, it's like a fourth tier without a number. 

Has the sentence below become untrue? 

> The FreeBSD Project differentiates platform targets into four tiers.

bsdimp:
Tier 4 is everything not supported. It seemed silly to number it when we revised the tier system 10ish years ago.

emaste:
LGTM

PR #189, opened 2023-06-17
Add OpenSearch status report

smortex:


website/content/en/status/report-2023-04-2023-06/opensearch.adoc
```diff
@@ -0,0 +1,14 @@
+=== OpenSearch
+
+Links: +
+link:https://opensearch.org/[OpenSearch] URL: link:https://opensearch.org/[] +
+
+Contact: OpenSearch Team <opensearch@FreeBSD.org>
+
+OpenSearch is a community-driven, open-source search and analytics suite derived from Apache 2.0 licensed ElasticSearch 7.10.2 & Kibana 7.10.2. It consists of a search engine daemon, OpenSearch, and a visualization and user interface, OpenSearch Dashboards. OpenSearch enables people to easily ingest, secure, search, aggregate, view, and analyze data. These capabilities are popular for use cases such as application search, log analytics, and more. With OpenSearch people benefit from having an open-source product they can use, modify, extend, monetize, and resell how they want.
```
lsalvadore, line 8: s/how they want/as they want/

website/content/en/status/report-2023-04-2023-06/opensearch.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+Since our last status report, the OpenSearch Team has regularly updated the OpenSearch and OpenSearch-Dashboards ports to follow upstream releases.
+
+While minor version updates are generally straightforward, some use case make this slightly harder and require more work. For this reason, the OpenSearch Team has recently decided to change the way we package OpenSearch.  Instead of our current scheme where we provide 2 ports, one for the latest major version (`textproc/openseach`, currently at version 2.7.0) and one for the legacy major version in maintenance mode(`textproc/openseach13`, currently at version 1.3.10), we will move to distribute each minor version as a new port (i.e. `textproc/openseach26`, `textproc/openseach27`, `textproc/openseach28` and so on).
```
lsalvadore, line 12: s/some use case/some use cases/

website/content/en/status/report-2023-04-2023-06/opensearch.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+Because upstream only provide updates for the latest major version, when a new minor version is made available and a new port is added, the previous minor version will be marked as deprecated, and the minor version before that will have an expiration date set, and will be removed at least three months later.  Users are advised to use the latest minor version, but this will not happen automatically and provide a grace period for users running software that needs to be updated accordingly.
```
lsalvadore, line 14: s/upstream only provide/upstream only provides/

s/this will not happen automatically and provide/this will not happen automatically and provides/

lsalvadore (changes requested):
Thanks for your report, I made a few inline comments.

Could you also please put one sentence per line? Thanks.

igalic (changes requested):
one sentence per line, please

smortex:
PR updated, thanks!

The OpenSearch team may still want to fine-tune the details of the changes described here, for this reason this PR is currently a draft.

website/content/en/status/report-2023-04-2023-06/opensearch.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+Since our last status report, the OpenSearch Team has regularly updated the OpenSearch and OpenSearch-Dashboards ports to follow upstream releases.
+
+While minor version updates are generally straightforward, some use cases make this slightly harder and require more work.
+For this reason, the OpenSearch Team has recently decided to change the way we package OpenSearch.
+Instead of our current scheme where we provide 2 ports, one for the latest major version (`textproc/openseach`, currently at version 2.7.0) and one for the legacy major version in maintenance mode(`textproc/openseach13`, currently at version 1.3.10), we will move to distribute each minor version as a new port (i.e. `textproc/openseach26`, `textproc/openseach27`, `textproc/openseach28` and so on).
+
+Because upstream only provides updates for the latest major version, when a new minor version is made available and a new port is added, the previous minor version will be marked as deprecated, and the minor version before that will have an expiration date set, and will be removed at least three months later.
```
pauamma, line 20: Do you mean "only provides updates for the latest minor version" here?

smortex: Hum, yeah this is ambiguous…

Let me describe this with an actual example:

As of today, `1.3.10` is the last 1.x major version, and is in maintenance mode, so the next versions will be `1.3.11`, `1.3.12`, etc and no `1.4.0` will ever be released.

`2.8.0` is the last 2.x major version. The next version *should* be `2.9.0`, but if some critical security issue is found, a `2.8.1` can be released quickly to just fix the security issue.  Then `2.9.0` will follow according to the initial schedule (unless a new critical security issue is found and fixed in `2.8.2`).  Once `2.9.0` is released, no more `2.8.x` will be released.

I tried to rephrase the sentence but it still seems very convoluted.  English not being a native language, I guess it does not help me :smile: But I hope it can help to iterate! 

```suggestion
Upstream ship new OpenSearch releases according to a public schedule.
Only the latest release of each major version is supported.
In case a critical security issue is found, it will be patched in a new patch release of that latest release of the affected major versions.

In order to ease migration, when a new minor version of OpenSearch is released and a new port is added, the port of the previous minor version will be marked deprecated, and the port of the second-to-previous minor version will have an expired date set to 3 months later.
```

pauamma: > As of today, `1.3.10` is the last 1.x major version, and is in maintenance mode, so the next versions will be `1.3.11`, `1.3.12`, etc and no `1.4.0` will ever be released.
> 
> `2.8.0` is the last 2.x major version. The next version _should_ be `2.9.0`, but if some critical security issue is found, a `2.8.1` can be released quickly to just fix the security issue. Then `2.9.0` will follow according to the initial schedule (unless a new critical security issue is found and fixed in `2.8.2`). Once `2.9.0` is released, no more `2.8.x` will be released.

OK, this sounds like the OpenSearch major version IDs have two components (`1.3` or `2.8`), not just one (`1` or `2`) as I thought. Is that correct?

smortex: They are supposed to follow [semver](https://semver.org/), so I think it makes more sense to stick to the semver MAJOR.MINOR.PATCH scheme.

MAJOR version 1 is in maintenance "mode", so no new feature will get in and the MINOR component will not change, only new PATCH release will happen.

MAJOR version 2 is the current branch where new features get in and MINOR is likely to be bumped when new releases are made.  We should generally see MAJOR.MINOR.0, MAJOR.MINOR.1 can happen but is unlikely while version 2 is the current one.

When version 3 will be released, version 2 will switch to maintenance mode where the PATCH component of the latest 2.MINOR.PATCH will be increased and MINOR will stay the same because no new features will come in, only bugfixes.

pauamma (approved):
I think I get it now, thanks.

lsalvadore (changes requested):

website/content/en/status/report-2023-04-2023-06/opensearch.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+Since our last status report, the OpenSearch Team has regularly updated the OpenSearch and OpenSearch-Dashboards ports to follow upstream releases.
+
+While minor version updates are generally straightforward, some use cases make this slightly harder and require more work.
+For this reason, the OpenSearch Team has recently decided to change the way we package OpenSearch.
+Instead of our current scheme where we provide 2 ports, one for the latest major version (package:textproc/opensearch[], currently at version 2.8.0) and one for the legacy major version in maintenance mode(package:textproc/opensearch13[], currently at version 1.3.10), we will move to distribute each minor version as a new port (i.e. package:textproc/opensearch26[], package:textproc/opensearch27[], package:textproc/opensearch28[] and so on).
+
+Because upstream only provides updates for the latest major version, when a new minor version is made available and a new port is added, the previous minor version will be marked as deprecated, and the minor version before that will have an expiration date set, and will be removed at least three months later.
```
smortex, line 20: @alonsobsd reading https://github.com/opensearch-freebsd/freebsd-ports/pull/1#issuecomment-1597776695 I understand that this is not really necessary and we can stick with the current scheme unless AWS does :hankey: and then we would handle this with a new version-specific port?

If so, this whole PR does not have much to say to our users, and I think we can dismiss it.  What do you think?

lsalvadore:
@smortex @alonsobsd : Can you please confirm if you want to submit this report, and in case you do if it is ready for commit? Thanks.

smortex:
It looks like nobody took position in favor of a direction or another.  IMHO this is status-quo, and therefore there is no need for a status report ATM.  I am closing it.

Or course, if needed, we can reopen this topic in the future :wink:.

Thank you all!

PR #223, opened 2023-07-15
Documentation: primer: AsciiDoc link macro; and Asciidoctor

grahamperrin:
Move towards proper use of the macro.

Prefixing: 

- only where necessary
- not with https: or mailto: URL schemes.

Whilst here, lowercase d in Asciidoctor.


grahamperrin:
@mojavelinux – primary maintainer of the language – is emphatic. The `link:` prefix is **superfluous if the target is a URL**.

lsalvadore:
I think this has already been discussed and it was decided to keep the `link:` prefix as it helps for translations in languages with non-latin fonts. But I might be wrong. I think @sergio-carlavilla might also remember something about it. And @dbaio too might be aware of some issues with translations.

documentation/content/en/books/fdp-primer/asciidoctor-primer/_index.adoc
```diff
@@ -240,8 +232,13 @@ And this would be rendered as.
 +
 extref:{ipsec-must}[IPSec-Must article]
 
+[[asciidoctor-link-macro]]
+=== The link macro
+
+This macro is not often needed. Use cases are https://docs.asciidoctor.org/asciidoc/latest/macros/link-macro/#when-to-use-the-link-macro[specified in Asciidoctor Docs].
```
grahamperrin, line 238: Amongst the few use cases: 

- The target must be enclosed in a passthrough to escape characters with special meaning

<https://github.com/freebsd/freebsd-doc/pull/223#issuecomment-1636760604>

> … languages with non-latin …

Would passthrough apply? 

Under **Escape and Prevent Substitutions**, [Passthroughs](https://docs.asciidoctor.org/asciidoc/latest/subs/prevent/#passthroughs) begins: 

> A passthrough is the primary mechanism by which to escape content in AsciiDoc. …

HTH

sergio-carlavilla:
We need to use the link macro, English is not the only language we're using in the documentation.
Please don't touch this warning.

grahamperrin:
What can be done to help readers understand the need?

TIA

grahamperrin:
https://github.com/grahamperrin/freebsd-doc/commit/5b9dea2e5a8d0857385fe8408b8c2d4b9caca975

> … URLlinkhttps://

– or, as a truly typed that part of the message: 

> `… URL:link:https://`

sergio-carlavilla:
Use: 
link:https://community.kde.org/FreeBSD[FreeBSD -- KDE Community Wiki]

Instead of:
link:https://community.kde.org/FreeBSD[FreeBSD -- KDE Community Wiki] URL: link:https://community.kde.org/FreeBSD[]

Why put the same two times?

grahamperrin:
> … Why put the same two times?

If I recall correctly, this duplication of sorts: 

- is documented somewhere
- is not related to the `link` macro
- relates to post-publication, post-conversion to plain text, distribution via email – `text/plain` specifically _without_ `text/html`.

If not already properly documented, then it should be fairly easy to find relevant commentary (and then document the logic) … probably in the https://github.com/freebsd/freebsd-quarterly archive.

sergio-carlavilla:
@lsalvadore do you need to put that two times for something related to the "plain" format?

About the PR by itselft, if you want to replace AsciiDoctor with Asciidoctor go ahead, np, but don't remove the link: macro, we need to put it to make the life more easier for translators. In some languages Asciidoctor don't recognize the links without the link macro

grahamperrin:
Can we be less insistent about blanket application of the `link:` prefix? 

At least: for content that we know _will not be translated_. 

----

In addition to the endless frustration caused by making things more difficult to proofread – and the increased likelihood of errors before and after proofreading – there is (at least, for me) the frustration of not knowing _exactly why_ the `link:` prefix is required in some, not all, cases. 

It will help to have a concrete example of wrongness as a result of not applying the prefix. Can someone point us towards an affected commit? If not a commit, then might something visible in Weblate (I'm not a user) help to understand the consequences?

TIA

sergio-carlavilla:
No, it's easier to say apply everywhere than to allow places where it doesn't.

Regarding the fact that there are things that are not going to be translated, I do not agree, all the documentation can and should be translated, English is not the only language that our users have.

The reason, Asciidoctor in some languages such as Japanese or Korean (languages that we have in our documentation) need to have the macro to know that what is being indicated is a link and not a normal text. It's not something related to weblate, it's something related to Asciidoctor. I know that in your documentation you say that this does not happen, but theory is one thing and practice is another.

Right now I can't spend time looking for an example in Japanese or Korean where this happens, but I remember it being like this when I did the conversion from Docbook to AsciiDoc and had to redo the scripts that parsed all the content.

lsalvadore:
> @lsalvadore do you need to put that two times for something related to the "plain" format?

Yes, we need that repetition for the email format of the status reports.

sergio-carlavilla:
Ok, so keep it, np
You can commit this change with only the replace of AsciiDoctor to Asciidoctor

sergio-carlavilla:
Upgrade the commit to see what you're gonna commit please

grahamperrin:
`% git -C /usr/doc show`

```patch
commit 5095f17a17430c9aeb130fee5afc42b66d0306be (HEAD -> main)
Author: Graham Perrin <grahamperrin@FreeBSD.org>
Date:   Sun Jul 16 11:54:08 2023 +0100

    Documentation: primer: Asciidoctor
    
    Asciidoctor (not AsciiDoctor).
    
    Reviewed-by:  carlavilla, salvadore
    Approved-by:  carlavilla
    Pull-request: https://github.com/freebsd/freebsd-doc/pull/223

diff --git a/documentation/content/en/books/fdp-primer/asciidoctor-primer/_index.adoc b/documentation/content/en/books/fdp-primer/asciidoctor-primer/_index.adoc
index 3ad0e15b60..9638651e68 100644
--- a/documentation/content/en/books/fdp-primer/asciidoctor-primer/_index.adoc
+++ b/documentation/content/en/books/fdp-primer/asciidoctor-primer/_index.adoc
@@ -1,16 +1,16 @@
 ---
-title: Chapter 6. AsciiDoctor Primer
+title: Chapter 6. Asciidoctor Primer
 prev: books/fdp-primer/doc-build
 next: books/fdp-primer/rosetta
-description: A brief introduction to AsciiDoctor
-tags: ["AsciiDoc", "AsciiDoctor", "Primer", "Introduction", "Guide"]
+description: A brief introduction to Asciidoctor
+tags: ["AsciiDoc", "Asciidoctor", "Primer", "Introduction", "Guide"]
 showBookMenu: true
 weight: 7
 path: "/books/fdp-primer/"
 ---
 
 [[asciidoctor-primer]]
-= AsciiDoctor Primer
+= Asciidoctor Primer
 :doctype: book
 :toc: macro
 :toclevels: 1
@@ -49,7 +49,7 @@ endif::[]
 
 Most FDP documentation is written with AsciiDoc.
 This chapter explains what that means, how to read and understand the documentation source, and the techniques used.
-To get a complete reference of the AsciiDoctor capabilities please consult the link:https://docs.asciidoctor.org/home/[Asciidoctor documentation].
+To get a complete reference of the Asciidoctor capabilities please consult the link:https://docs.asciidoctor.org/home/[Asciidoctor documentation].
 Some of the examples used in this chapter have been taken from the link:https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference[AsciiDoc Syntax Quick Reference].
 
 [[asciidoctor-primer-overview]]
@@ -97,7 +97,7 @@ To remove */tmp/foo*, use man:rm[1].
 [[asciidoctor-headings]]
 == Headings
 
-AsciiDoctor supports six headings levels.
+Asciidoctor supports six headings levels.
 If the document type is `article` only one level 0 (`=`) can be used.
 If the document type is `book` then there can be multiple level 0 (`=`) headings.
 
@@ -155,7 +155,7 @@ And this is the second paragraph.
 [[asciidoctor-lists]]
 == Lists
 
-AsciiDoctor supports a few types of lists, the most common are `ordered` and `unordered`.
+Asciidoctor supports a few types of lists, the most common are `ordered` and `unordered`.
 To get more information about lists, see link:https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#lists[AsciiDoc Syntax Quick Reference].
 
 [[asciidoctor-ordered-lists]]
@@ -214,14 +214,14 @@ link:https://www.FreeBSD.org[FreeBSD]
 
 [NOTE]
 ====
-As the AsciiDoctor documentation describes, the `link` macro is not required when the target starts with a URL scheme like `https`.
-However, it is a good practice to do this anyway to ensure that AsciiDoctor renders the link correctly, especially in non-latin languages like Japanese.
+As the Asciidoctor documentation describes, the `link` macro is not required when the target starts with a URL scheme like `https`.
+However, it is a good practice to do this anyway to ensure that Asciidoctor renders the link correctly, especially in non-latin languages like Japanese.
 ====
 
 [[asciidoctor-links-internal]]
 === Internal link
 
-To point to another book or article the AsciiDoctor variables should be used.
+To point to another book or article the Asciidoctor variables should be used.
 For example, if we are in the `cups` article and we want to point to `ipsec-must` these steps should be used.
 
 . Include the [.filename]#urls.adoc# file from [.filename]#~/doc/shared# folder.
@@ -230,7 +230,7 @@ For example, if we are in the `cups` article and we want to point to `ipsec-must
 \include::shared/{lang}/urls.adoc[]
 ....
 +
-. Then create a link using the AsciiDoctor variable to the `ipsec-must` article.
+. Then create a link using the Asciidoctor variable to the `ipsec-must` article.
 +
 ....
 extref:{ipsec-must}[IPSec-Must article]
@@ -243,5 +243,5 @@ extref:{ipsec-must}[IPSec-Must article]
 [[asciidoctor-conclusion]]
 == Conclusion
 
-This is the conclusion of this AsciiDoctor primer.
+This is the conclusion of this Asciidoctor primer.
 For reasons of space and complexity, several things have not been covered in depth (or at all).
```

sergio-carlavilla:
LGTM

grahamperrin:
Closing here, thanks. 

<https://github.com/freebsd/freebsd-doc/pull/223#issuecomment-1637003380>

> … an example in Japanese or Korean …

@mojavelinux visualised a theoretical example, which helped to jog my memory. Then: 

- `rg "link:http" /usr/doc/documentation/content/ja`
- `rg "link:http" /usr/doc/website/content/ja`

– eventually, I found a handful of real cases in FreeBSD areas. 

At a glance, **these cases are rare** … one or two under <https://docs.freebsd.org/ja/books/porters-handbook/porting-dads/#dads-intro>: 

image

- note what's highlighted (a search result), the extent of the underlined link, and what's to the left of the underline.

### Essentially

Beware of a URL macro adjacent to a **non-space** character.

----

### With spaces (non-adjacent)

Areas such the two below might be the result of confusion: 

`それから "link:http://www.FreeBSD.org/cgi/man.cgi?su[su]" を使って `root` になってください。 `

image

And so on.

sergio-carlavilla:
I remove myself from this PR, the amount of noise that is generated sometimes by things *so* trivial is terrible, having other much more important things to work on

grahamperrin:
No less terrible: 

> the endless frustration caused by making things more difficult to proofread

If that's _so_ trivial in the eyes of others, I'm in the wrong project. 

Enough said.

PR #232, opened 2023-07-25
ZFS: fix a typo

soobinrho:
Managing Clones

https://docs.freebsd.org/en/books/handbook/zfs/

'Creating a clone makes it an exact copy of the state the dataset as it was when taking the snapshot.'

– is fixed to 'the state the dataset was in'.


igalic (approved):

sergio-carlavilla:
Fixed in https://cgit.freebsd.org/doc/commit/?id=2cfef2379a95eaef096a75eabe1d110185e355a5
Thanks!

PR #236, opened 2023-07-27
Update ifapi.adoc

grahamperrin:
https://www.freebsd.org/status/report-2023-04-2023-06/#_network_interface_api_ifapi

- link: prefix is visible
- the result is not a link.

Fix breakage that would not have occurred without the contentious link: prefix

### DIFF
--- a/website/content/en/status/report-2023-04-2023-06/ifapi.adoc
+++ b/website/content/en/status/report-2023-04-2023-06/ifapi.adoc
@@ -1,7 +1,7 @@
 === Network Interface API (IfAPI)
 
-Links: +
-link:https://wiki.freebsd.org/projects/ifnet[Original project page] URL: link:https://wiki.freebsd.org/projects/ifnet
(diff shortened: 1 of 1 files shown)


website/content/en/status/report-2023-04-2023-06/ifapi.adoc
```diff
@@ -1,7 +1,7 @@
 === Network Interface API (IfAPI)
 
-Links: +
-link:https://wiki.freebsd.org/projects/ifnet[Original project page] URL: link:https://wiki.freebsd.org/projects/ifnet
+Link: +
```
lsalvadore, line 3: I think we usually entitle the section "Links" and I would keep it as usual even if there is just one link, but if you prefer to switch to singular it is fine.

lsalvadore (approved):

sergio-carlavilla:
Merged.

PR #237, opened 2023-07-27
Update simd.adoc

grahamperrin:
https://www.freebsd.org/status/report-2023-04-2023-06/#_simd_enhancements_for_amd64

- link: prefix is visible
- the result is not a link

Fix breakage that would not have occurred without the contentious link: prefix


lsalvadore (approved):

sergio-carlavilla:
Merged.

PR #242, opened 2023-07-27
Update pf.adoc

grahamperrin:
https://www.freebsd.org/status/report-2023-04-2023-06/#_pf_improvements

~~Subheadings that begin 'pfsync' are true to lowercase.~~ 

~~Be true for the primary subheading that begins 'pf'.~~

Fix the lines for the three sponsors (there should be three lines, not a single line for all sponsors).


website/content/en/status/report-2023-04-2023-06/pf.adoc
```diff
@@ -1,4 +1,4 @@
-=== Pf Improvements
+=== pf Improvements
```
lsalvadore, line 1: This has already been discussed with the report submitters and they choose to capitalize Pf but not pfsync, so please leave it as it is.

lsalvadore (approved):
Approved only the sponsors change (see inline comment).

sergio-carlavilla:
Merged.

PR #250, opened 2023-08-27
Fix broken link in porters-handbook

dzvon:
Currently the `ports-using` macro is rendered as https://docs.freebsd.org/en/books/handbook#ports-using, which doesn't look right.

This commit changes the rendered link to https://docs.freebsd.org/en/books/handbook/ports#ports-using.


sergio-carlavilla:
Yep, good catch :)

dzvon:
@sergio-carlavilla I'm learning FreeBSD recently, reading documents like crazy :D

sergio-carlavilla:
Fixed in: https://cgit.freebsd.org/doc/commit/?id=e971ace59e7171439c936e4f6be5e5aec24d5f53

Thanks :)

PR #262, opened 2023-09-28
Fix typo in VNET Jail configuration example

kenrap:
I found this typo which caught me off guard with a confusing error when trying to create a VNET Jail for the first time.

I thought I would spend a few minutes to make a quick fix to send upstream. 👍


sergio-carlavilla:
Fixed in https://cgit.freebsd.org/doc/commit/?id=1c72508eb7b4f2a02f15b1400cf3678abb518a77
Thanks!

PR #264, opened 2023-09-28
Handbook - Security: document OpenSSL's FIPS provider

khorben:
The new FIPS provider module from OpenSSL 3 is protected from use without passing an integrity check, which is expected to be configured by the local system administrator.

This new section documents a way to configure this on FreeBSD, so that every user on the system is allowed to pass this check and use the FIPS module.


sergio-carlavilla:
Uo! Thanks for that!
I’ll make the commit

khorben:
> Uo! Thanks for that! I’ll make the commit

Hi Sergio! Do you need any additional input from my side?

sergio-carlavilla:
Done in: https://cgit.freebsd.org/doc/commit/?id=fb21a60804dfd1cfbb8f00afc4439bb3e5001cca
Sorry for the delay, if this happens again, please ping me by email.

PR #269, opened 2023-09-29
Status: 2023q3: SIMD: correction, specificity

grahamperrin:
Markup: a hard line break between the first and second links.

Manual page: link to the specified section (7).

Use 'either …, or …' within the second of four list items.

### DIFF
--- a/website/content/en/status/report-2023-07-2023-09/simd.adoc
+++ b/website/content/en/status/report-2023-07-2023-09/simd.adoc
@@ -1,8 +1,8 @@
 === SIMD enhancements for amd64
 
 Links: +
-link:http://fuz.su/~fuz/freebsd/2023-04-05_libc-proposal.txt[Project proposal] URL: link:http://fuz.su/~fuz/freebsd/2023-04-05_libc-proposal.txt[]
-link:https://man.freebsd.org/cgi/man.cgi?query=simd&manpath=FreeBSD+15.0-CURRENT[simd(7)] URL: link:https://man.freebsd.org/cgi/man.cgi?query=simd&manpath=FreeBSD+15.0-CURRENT[]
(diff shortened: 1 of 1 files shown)


website/content/en/status/report-2023-07-2023-09/simd.adoc
```diff
@@ -12,7 +12,7 @@ The goal of this project is to provide SIMD-enhanced versions of common libc fun
 For each function optimised, up to four implementations will be provided:
 
  * a *scalar* implementation optimised for amd64, but without any SIMD usage,
- * a *baseline* implementation using SSE and SSE2 or alternatively an *x86-64-v2* implementation using all SSE extensions up to SSE4.2,
+ * either a *baseline* implementation using SSE and SSE2, or an *x86-64-v2* implementation using all SSE extensions up to SSE4.2,
```
grahamperrin, line 15: This may be slightly easier to read in, for example, the plain text email.

grahamperrin:
@clausecker for your consideration. Thanks.

clausecker:
I used this phrasing because for all but two functions, it's likely going to be a baseline implementation and x86-64-v2 is a bit of a special case.  But if this wording feels awkward to a native speaker, sure, change it.

I don't think we need a `sektion=7` query parameter in the URL as there is no `simd` man page in any other section, but I guess it won't hurt.

LGTM.

grahamperrin:
@clausecker thanks. 

Subtly: if `sektion=7` is omitted, then the title of the linked page will be _simd_, which is relatively vague for title bar, tab, bookmarking and other purposes.

An 'or alternatively' always catches my eye. Whilst grammatically correct, it's sometimes a trigger to look at surrounding texts.

lsalvadore:
Merged, thanks.

PR #294, opened 2023-11-11
Update NFS locking instructions

chrislongros:


igalic (commented):
I'm somewhat confused why this Greek handbook page is English

PR #303, opened 2023-12-07
bsdinstall: modernize the screenshots

khorben:
This contains a new version of every screenshot still relevant to the bsdinstall section of the FreeBSD handbook. They were made after booting in EFI mode instead of the traditional BIOS mode, for a more modern look as well as a more accurate resemblance to the actual look & feel of the installer on current hardware.

Sponsored by:	The FreeBSD Foundation


bsdjhb (approved):
All look good to me.  Easier to review here than on phabricator since it shows the image previews.

khorben:
> All look good to me. Easier to review here than on phabricator since it shows the image previews.

Thank you! I thought so too :)

emaste:
Looks good. A couple of differences I noticed:

- No ACPI option on bootloader screen now
- Manual configuration -- No is now highlighted
- GPT partition type is now selected
- Root password - screen grab while entering the password the first time rather than the retype

khorben:
> Looks good. A couple of differences I noticed:
> 
> * No ACPI option on bootloader screen now
> * Manual configuration -- No is now highlighted
> * GPT partition type is now selected
> * Root password - screen grab while entering the password the first time rather than the retype

Regarding ACPI, according to https://www.freebsd.org/releases/13.0R/hardware/ "FreeBSD/amd64 requires [ACPI] support," which must be the reason the option is gone from the bootloader options menu. Since i386 support is planned to be obsoleted in 15.0, should I add a commit mentioning something like "(if ACPI support is present and can be disabled)"? (e.g., line 311 of `documentation/content/en/books/handbook/bsdinstall/_index.adoc`)

For the manual configuration, according to the text, either answer can be illustrated in the screenshots. Ideally no manual configuration is needed ;)

For GPT, I think it is more appropriate to illustrate this instead of MBR for modern setups.

For the root password you are right, I can push a new screenshot to illustrate the whole process.

dbaio:
- The resolution of some images is slightly larger than the previous ones, but you maintained a consistent pattern across all of them, which is good.

- The overall size of the images is smaller, which is nice.

- It looks ok on mobile view.

- It looks ok in the PDF.

Thanks for your work.

I'll push this through.

PR #310, opened 2023-12-21
freebsd-src-lsp/_index.adoc: write "keys" instead of "key"

rilysh:
It should be
>  Entries with duplicated command key**s** are merged.

Webpage: https://docs.freebsd.org/en/articles/freebsd-src-lsp/#_usage_2


chrislongros (approved):
:+1: thanks for noticing

igalic (approved):

dbaio:
Committed on 139e052d84e69ec418cb5fdff037354ab948e3f4. Thanks for your contribution.

PR #318, opened 2024-01-02
handbook (Desktop Environments): add zathura

sonphantrung:
[zathura](https://pwmt.org/projects/zathura/) is a neat and lightweight document viewer that I use, and I want to add it to the handbook ;).


igalic (changes requested):
this text looks good overall, but i didn't dwell too deeply on it

here's some structural remarks:

- this doesn't need to be 2 commits, please squash them
- we don't accept patches from noreply email addresses, please change that

PR #327, opened 2024-01-12
wireless: changed wlans_iwn0 to wlans_iwm0

confusedjpg:
Just fixing a small typo (that I noticed after going insane).

### DIFF
--- a/documentation/content/en/books/handbook/network/_index.adoc
+++ b/documentation/content/en/books/handbook/network/_index.adoc
@@ -600,7 +600,7 @@ network={
 
 [source,shell]
 ....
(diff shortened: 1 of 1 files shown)


igalic:
[iwn](https://man.freebsd.org/iwn(4)) which is a real thing, referenced earlier, https://docs.freebsd.org/en/books/handbook/network/#config-identify-network-adapter

confusedjpg:
Here I go not being careful enough again.. 
I see, thanks for the correction and I apologize for this then.

confusedjpg:
Fair enough, not sure if this one is more common as I didn't have to deal with this in the past, but I get the idea.

PR #326, opened 2024-01-10
Typo in DNS

ZePotente:
understanded -> understood

PR #328, opened 2024-01-13
Update _index.adoc

AndrShat:
I've corrected several bugs in example of thin jail based on nullfs to make it work.

### DIFF
--- a/documentation/content/en/books/handbook/jails/_index.adoc
+++ b/documentation/content/en/books/handbook/jails/_index.adoc
@@ -598,7 +598,7 @@ In case of using UFS the man:cp[1] program can be used by executing the followin
 [source,shell]
 ....
 # mkdir /usr/local/jails/containers/thinjail
-# cp -R /usr/local/jails/templates/13.2-RELEASE-skeleton /usr/local/jails/containers/thinjail
+# cp -R /usr/local/jails/templates/13.2-RELEASE-skeleton/* /usr/local/jails/containers/thinjail
 ....
 
 Then create the directory in which the base template and the skeleton will be mounted:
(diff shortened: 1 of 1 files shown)


rilysh:
@AndrShat I think you should also add a valid email address in your commit. See #326

documentation/content/en/books/handbook/jails/_index.adoc
```diff
@@ -598,7 +598,7 @@ In case of using UFS the man:cp[1] program can be used by executing the followin
 [source,shell]
 ....
 # mkdir /usr/local/jails/containers/thinjail
-# cp -R /usr/local/jails/templates/13.2-RELEASE-skeleton /usr/local/jails/containers/thinjail
+# cp -R /usr/local/jails/templates/13.2-RELEASE-skeleton/* /usr/local/jails/containers/thinjail
```
dbaio, line 601: what was wrong here?

documentation/content/en/books/handbook/jails/_index.adoc
```diff
@@ -636,6 +636,13 @@ thinjail {
 }
 ....
 
+The name of a file in [.filename]#jail.conf.d# must be thinjail.conf. Otherwise you have to create also [.filename]#/etc/jail.conf#:
```
dbaio, line 639: You can add the reason here, such as the requirement for the file name to match the jail name, in this case, thinjail.conf.

dbaio (changes requested):

documentation/content/en/books/handbook/jails/_index.adoc
dbaio: Ahh, I got it.. the directory is already created; ignore this.

PR #333, opened 2024-01-25
dev-handbook/x86: remove line numbers from the example

rilysh:
These line numbers seem to be mistakenly copied from the editor. Since the user is (likely) going to copy-and-paste the example, we should remove these line numbers.

### DIFF
--- a/documentation/content/en/books/developers-handbook/x86/_index.adoc
+++ b/documentation/content/en/books/developers-handbook/x86/_index.adoc
@@ -541,22 +541,22 @@ We are now ready for our first program, the mandatory Hello, World!
 
 [.programlisting]
 ....
-1:	%include	'system.inc'
- 2:
- 3:	section	.data
- 4:	hello	db	'Hello, World!', 0Ah
- 5:	hbytes	equ	$-hello
- 6:
- 7:	section	.text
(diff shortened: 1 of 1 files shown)


igalic (approved):

pauamma:
I'm on the fence about this. On one hand you're correct that copypasting is tempting. On the other hand this would make it harder to follow the explanation below, which refers to line numbers, and section A.6.1 (https://docs.freebsd.org/en/books/developers-handbook/x86/#x86-assemble-1) would need a change as well, for consistency.

igalic:
I'm surprised ou rasciidoc doesn't provide us with an option for numbering lines of `programlisting`s

rilysh:
> I'm on the fence about this. On one hand you're correct that copypasting is tempting. On the other hand this would make it harder to follow the explanation below, which refers to line numbers, and section A.6.1 (https://docs.freebsd.org/en/books/developers-handbook/x86/#x86-assemble-1) would need a change as well, for consistency.

I've removed the" (except the line numbers)" part, and now it should be good.

dbaio:
Committed on ec09b757de30f0a7c5d6b61d05933be8a883ebb3
Thank you!

PR #336, opened 2024-02-07
Added a clarification about jail configuration

oliverepper:
- Files in /etc/jail.conf.d must have a .conf extension in order for the tools to pick them up.
- Autostart of jails requires the /etc/jail.conf file.

### DIFF
--- a/documentation/content/en/books/handbook/jails/_index.adoc
+++ b/documentation/content/en/books/handbook/jails/_index.adoc
@@ -250,9 +250,6 @@ There are two ways to configure jails.
 The first one is to add an entry for each jail to the file [.filename]#/etc/jail.conf#.
 The other option is to create a file for each jail in the directory [.filename]#/etc/jail.conf.d/#.
 
-There is no right or wrong option.
-Each administrator must choose the one that best suits their needs.
-
 In case a host system has few jails, an entry for each jail can be added in the file [.filename]#/etc/jail.conf#.
 If the host system has many jails, it is good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
 
--- a/documentation/content/en/books/handbook/jails/_index.adoc
+++ b/documentation/content/en/books/handbook/jails/_index.adoc
@@ -256,6 +256,15 @@ Each administrator must choose the one that best suits their needs.
 In case a host system has few jails, an entry for each jail can be added in the file [.filename]#/etc/jail.conf#.
 If the host system has many jails, it is good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
 
+The files you put in the [.filename]#/etc/jail.conf.d/# directory must have `.conf` as their extension.
+
+If you enabled jails in [.filename]#/etc/rc.conf# you should put the following into [.filename]#/etc/jail.conf#:
+
+[.programmlisting]
+....
+.include "/etc/jail.conf.d/*.conf";
+....
+
 A typical jail entry would look like this:
 
 [.programlisting]
--- a/documentation/content/en/books/handbook/jails/_index.adoc
+++ b/documentation/content/en/books/handbook/jails/_index.adoc
@@ -251,7 +251,7 @@ The first one is to add an entry for each jail to the file [.filename]#/etc/jail
 The other option is to create a file for each jail in the directory [.filename]#/etc/jail.conf.d/#.
 
 In case a host system has few jails, an entry for each jail can be added in the file [.filename]#/etc/jail.conf#.
-If the host system has many jails, it is good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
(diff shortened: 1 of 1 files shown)


documentation/content/en/books/handbook/jails/_index.adoc
```diff
@@ -250,12 +250,18 @@ There are two ways to configure jails.
 The first one is to add an entry for each jail to the file [.filename]#/etc/jail.conf#.
 The other option is to create a file for each jail in the directory [.filename]#/etc/jail.conf.d/#.
 
-There is no right or wrong option.
-Each administrator must choose the one that best suits their needs.
-
 In case a host system has few jails, an entry for each jail can be added in the file [.filename]#/etc/jail.conf#.
 If the host system has many jails, it is good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
```
pauamma, line 254:
```suggestion
If the host system has many jails, it is a good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
```

pauamma (approved):

documentation/content/en/books/handbook/jails/_index.adoc
oliverepper: Thanks.

pauamma (approved):

documentation/content/en/books/handbook/jails/_index.adoc
```diff
@@ -250,11 +250,17 @@ There are two ways to configure jails.
...
-There is no right or wrong option.
-Each administrator must choose the one that best suits their needs.
-
 In case a host system has few jails, an entry for each jail can be added in the file [.filename]#/etc/jail.conf#.
-If the host system has many jails, it is good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
+If the host system has many jails, it is a good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
+
+The files you put in the [.filename]#/etc/jail.conf.d/# directory must have `.conf` as their extension.
+
+If you enabled jails in [.filename]#/etc/rc.conf# you should put the following into [.filename]#/etc/jail.conf#:
```
dbaio, line 258: "While it's not forbidden to use 'you', it might be better to use it less in this context."
See https://docs.freebsd.org/en/books/fdp-primer/book/#writing-style-be-clear

oliverepper: Thanks a lot for the hint. I wasn't aware of that document. Cool. I will fix the text.

documentation/content/en/books/handbook/jails/_index.adoc
```diff
@@ -250,11 +250,15 @@ There are two ways to configure jails.
...
 In case a host system has few jails, an entry for each jail can be added in the file [.filename]#/etc/jail.conf#.
-If the host system has many jails, it is good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
+If the host system has many jails, it is a good idea to have one configuration file for each jail in the [.filename]#/etc/jail.conf.d/# directory.
+
+The files in [.filename]#/etc/jail.conf.d/# must have `.conf` as their extension and have to be included in [.filename]#/etc/jail.conf#:
+
+[.programmlisting]
+....
+.include "/etc/jail.conf.d/*.conf";
+....
```
dbaio, line 261: I believe this phrase is outdated, especially considering FreeBSD versions 13 and 14.

See https://github.com/freebsd/freebsd-src/blob/main/libexec/rc/rc.d/jail#L122-L130

Perhaps consider changing it to something like this:
```
The files in [.filename]#/etc/jail.conf.d/# must be named as `jailname.conf`.
```

There is no need to use that include anymore.

oliverepper: Not sure if I really understand you right. I double tested both, now. The files in `jail.conf.d` do have to have the extension (see screenshot) and if there is no /etc/jail.conf or just an empty one (without the .include statement) then jails configured in `jail.conf.d` will not autostart. (second screenshot). If I touch `/etc/jail.conf` the jail will not be started. So I do still believe that the extension and the include is needed. Just tested everything again to double check.

Bildschirmfoto 2024-02-26 um 09 15 16

Bildschirmfoto 2024-02-26 um 09 18 40

oliverepper: I see. I'd personally prefer the include because it appears to be more flexibel and less typing. I'd still like to update the handbook, because the way it is now jails simply don't start. Neither at system start, nor via `service jail start`. Shall I make it clear that there are two variants for the autostart? Or is the `.include` variant officially deprecated? I would ignore that, then.

oliverepper: Thanks for your time reviewing this. Very much appreciated.

dbaio:
Committed on df4729f3c0e9dcc0660c1ae81baeb40b1cb46566
Thank you!

PR #304, opened 2023-12-08
network: update ifconfig wlan create

subpop:
Update the example ifconfig create command to use the correct parent device name parameter (wlandev instead of wlandevice).

PR #340, opened 2024-02-20
Change wlandevice -> wlandev.

jeannekamikaze:
As of `14.0`, `wlandev` appears to be the right parameter, not `wlandevice`:

https://man.freebsd.org/cgi/man.cgi?ifconfig

I configured my wifi card correctly following the instructions on the page using `wlandev`.

### DIFF
--- a/documentation/content/en/books/handbook/network/_index.adoc
+++ b/documentation/content/en/books/handbook/network/_index.adoc
@@ -619,7 +619,7 @@ To find out what wireless network cards are in the system check the section <<co
 
 [source,shell]
 ....
-# ifconfig wlan0 create wlandevice iwm0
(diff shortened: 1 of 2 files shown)


pauamma:
"wlandev" on 13.2 as well.

jeannekamikaze:
Sorry, is that a request for changes or just pointing out that it's also "wlandev" on 13.2? I grepped the whole tree for "wlandevice" and these seem to be the only two occurrences.

dbaio:
Hi, I'll close this PR because we are already handling this on another PR,  https://github.com/freebsd/freebsd-doc/pull/304/
and next time, don't need to change .po files.
Thank you!

PR #349, opened 2024-03-08
fix package URLs, and highlight the elisp code

rilysh:
1. There seem to be no package called vim-console or nano-devel in the ports tree, rather it's now just called vim or nano (depending on which one a user will install).

2. Highlight the emacs lisp syntax.


dbaio:
Committed on https://github.com/freebsd/freebsd-doc/commit/972d1017e7a25b11eb9be827acd2ff05778ca60e
Thank you!

PR #353, opened 2024-03-19
Update _index.adoc

KolusionWhatever:


KolusionWhatever:
What year did the third edition of the handbook come out? It needs updating in the heading.

KolusionWhatever:
And also, is the fourth edition the latest edition? And is it still relevant to even mention these older editions? It seems more like history than actual practical instructions. I think its better to keep documentation is minimal as possible because it makes it easier to navigate and less mental work. So I'm thinking cutout the older sections about history.

KolusionWhatever:
Knowing the year the fourth edition came out is also needed for the heading if following the scheme of including the year in the headings.

Personally, I feel it is pointless including the year in the heading as it doesn't serve any purpose, the same way I feel the first, second and third edition sections don't actually serve the user any purpose as they are irrelevant to the currently supported versions of FreeBSD which this documented is intended to cover.

PR #355, opened 2024-03-19
Update _index.adoc

KolusionWhatever:
Changed to exact package names for consistency throughout the document. Different names could confuse people.


KolusionWhatever:
"x11/mate" is under the section "Package", but "x11/mate" is equivalent to 'Port section/package', like described here:

https://ports.freebsd.org/cgi/ports.cgi


So that isn't logical.

PR #360, opened 2024-03-25
handbook/wine: add Mizuma's source code URL, and small typos

rilysh:
1. It's particularly interesting to know what games/apps are currently "supported" through Mizuma, so it's a good idea to link the source code URL (winetricks source code URL is also already mentioned.)
2. Remove the word "some" from the synopsis. It's ambiguous to know the meaning of "some" here, whereas Wine is intended (and does support) a wide range of Windows applications out of the box.
3. Use the word GitHub instead Github (ignorable).
4. Also edit the Arch Wiki URL (it redirects to that new URL from the previous one) (forgot to mention in the commit).

### DIFF
--- a/documentation/content/en/books/handbook/wine/_index.adoc
+++ b/documentation/content/en/books/handbook/wine/_index.adoc
@@ -52,7 +52,7 @@ endif::[]
 == Synopsis
 
 https://www.winehq.org/[WINE], which stands for Wine Is Not an Emulator, is technically a software translation layer.
-It enables to install and run some software written for Windows(R) on FreeBSD (and other) systems.
+It enables to install and run software written for Windows(R) on FreeBSD (and other) systems.
 
 It operates by intercepting system calls, or requests from the software to the operating system, and translating them from Windows(R) calls to calls that FreeBSD understands.
 It will also translate any responses as needed into what the Windows(R) software is expecting.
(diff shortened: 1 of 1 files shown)


sergio-carlavilla:
Thanks for the changes. I'll make the commit this night.

rilysh:
Friendly reminder, ping.

concussious (approved):

documentation/content/en/books/handbook/wine/_index.adoc
```diff
@@ -52,7 +52,7 @@ endif::[]
 == Synopsis
 
 https://www.winehq.org/[WINE], which stands for Wine Is Not an Emulator, is technically a software translation layer.
-It enables to install and run some software written for Windows(R) on FreeBSD (and other) systems.
+It enables to install and run software written for Windows(R) on FreeBSD (and other) systems.
```
kaduk, line 55: I fixed this with a follow-up commit since it was a preexisting issue unaffected by your edit, but "enables" is a transitive verb that needs a direct object, like "it enable users to install and run software" ... but "users" is pretty vague in this context so I went with an alternative verb that is intransitive.

documentation/content/en/books/handbook/wine/_index.adoc
```diff
@@ -556,7 +556,7 @@ image::winetricks-uninstall-3.png[]
 [[homura]]
 === Mizutamari
 
-Mizutamari is an application similar to `winetricks`, although it was inspired by the https://lutris.net/[Lutris] gaming system for Linux.
+https://codeberg.org/Alexander88207/Mizutamari[Mizutamari] is an application similar to `winetricks`, although it was inspired by the https://lutris.net/[Lutris] gaming system for Linux.
```
kaduk, line 559: It's also available in the ports collection as games/mizuma, but the link to its homepage seems more appropriate than the ports collection entry, for this purpose, so I left it alone.

kaduk (approved):
This is committed to main now, as 224f316cbd8de2669856ae298778853d45c693cc.
So we should close the PR.

bsdimp:
This is committed.

documentation/content/en/books/handbook/wine/_index.adoc
rilysh: Thanks for the update!

PR #363, opened 2024-03-29
status: 2024q1: FreshPorts: #history (link), longevity

grahamperrin:
```text
#history – not #packages – shows history.

Over 23 years since FreshPorts started.
```


grahamperrin:
Glancing at the three headings, their levels, and content: 

image

– should the first heading be simply _FreshPorts_?

@dvl

lsalvadore:
Merged, thanks.

I have kept the headers as they are. If the report author contacts us to change them I will make the change.

PR #371, opened 2024-04-13
fix jexec example

nt1:
Correct the usage of `service` in the jexec example.


kaduk:
It looks like this issue got resolved by maxim in commit 6ab40f7aa6c1b6c15f94eca491bbb4357935dfdf which was marked as fixing https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278341 but did not notice this PR.
Thank you for the submission!

bsdimp:
Closed per submitter request

PR #378, opened 2024-05-04
status: 2024q1: corrections

grahamperrin:
`Fix a link:https:// … URL: link:https: line.`


lsalvadore:
Thanks @grahamperrin, if you have any other fixes, please complete your draft as soon as you can: I would like to send the reports by e-mail soon enough. As you know, we are already a bit late...

lsalvadore:
I have merged this single change for now. I am also going to send the e-mail version of the reports within today.

If you find other issues, feel free to submit them in a new pull request. They will be too late for the e-mail version but still good for the web version.

Thanks.

grahamperrin:
Thanks. Sorry that I'm further away than I was in the past. I caught just the one thing, at a glance, I looked first at the installer section because it was (naturally) a focus of attention from news sites and so on.

PR #382, opened 2024-05-21
Fix thinjail instructions on UFS

egwynn:
### Description 

The zfs instructions say to clone a snapshot from the skeleton to the new thinjail. But the UFS instructions say to _make_ the thinjail directory first and then copy the skeleton into it. This will cause your new jail to be inside of `/usr/local/jails/containers/thinjail/13.2-RELEASE-skeleton` instead of just `/usr/local/jails/containers/thinjail`.


dbaio:
Committed on a607e18d9984dc9b36f376dec15321371f6ea93b
Thank you!

PR #383, opened 2024-05-24
Fix wording to reflect commands without DHCP

mario-campos:
The `sysrc` commands would be enabled when an interface configures its address via SLAAC, not DHCP. Hence, the instruction is incorrect.


rilysh (approved):

dbaio:
Committed on dc68d1c4655fdd937b8febf5aaa7afabd58d4823
Thank you both!

PR #396, opened 2024-07-14
fix(280150): 'Edit this page' link is broken in FreeBSD handbook

danielsoro:
I noticed that the links to the edit page are broken and saw the following bug reported: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=280150

This change fixes the creation of the edit URL to point to the correct file. 

The `IF` in the list template for the book was introduced because the load of the main page from the menu is loaded with the list template, but this main page doesn't present any specific section in the pathSections, which I named `sectionName`.

Hope it helps. :)

### DIFF
--- a/documentation/themes/beastie/layouts/articles/list.html
+++ b/documentation/themes/beastie/layouts/articles/list.html
@@ -47,9 +47,10 @@
           {{ $docType := index $pathSections 1 }}
           {{ $docName := index $pathSections 2 }}
           {{ $pdfFilename := printf "%s_%s.pdf" $docName $docLang }}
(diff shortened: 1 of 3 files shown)


msimerson:
I noticed this today, thanks for the PR. Hopefully it'll get merged. 🤞🏻

grahamperrin:
Does this result in an edit interface? 

That was the norm, in the past, but then it was decided (not by me) that edit should not edit.

sergio-carlavilla:
I'm gonna make the commit. Sorry for the big delay!
I had personal issues but I'm back!

sergio-carlavilla:
Done!

PR #399, opened 2024-07-26
modify shasum check command

lgrn:
by default, sha256sum and sha512sum will look for any file mentioned in the CHECKSUM file, which is all of them:
```
$ wc -l < CHECKSUM.SHA512-FreeBSD-14.1-RELEASE-amd64 10
```
with the current command, this will lead to 9 failures and 1 OK. the fact that you specify the filename to be checked does not seem to mean that is the *only* file that is checked.

for this reason, I suggest that the filename to be checked is removed, and replaced with the flag --ignore-missing which is present in both commands on both BSD and Linux.

this flag means that all files mentioned will be checked, just like before, but output will only be returned for successful checks. example:
```
$ sha512sum -c CHECKSUM.SHA512-FreeBSD-14.1-RELEASE-amd64 --ignore-missing
FreeBSD-14.1-RELEASE-amd64-memstick.img: OK
```
this assumes that the img file is in the same directory, but this was already assumed in the previous version.

### DIFF
--- a/documentation/content/en/books/handbook/bsdinstall/_index.adoc
+++ b/documentation/content/en/books/handbook/bsdinstall/_index.adoc
@@ -183,12 +183,12 @@ Note that you need to compare the calculated _checksum_ against the correct file
 FreeBSD provides man:sha256[1] and man:sha512[1] that can be used for calculating the _checksum_.
 Other operating systems have similar programs.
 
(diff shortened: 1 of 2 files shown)


concussious:
LGTM. I think just fix the author -- "All commits have, as the author, your name and valid email address as you would like to see them in the FreeBSD repository. Fake github.com addresses cannot be used." ~FreeBSD-src/CONTRIBUTING.md

lgrn:
Fixed author and email in c22c9c525648952105f9ab43948bdd53f885226c

PR #400, opened 2024-07-31
Update my PGP key (expiry date, new sign- and encrypt subkeys)

adriaandegroot:
I saw a question from gad@ on _developers_ and realised that my GPG key hadn't been updated in some time.


adriaandegroot:
(For cross-validation, also at https://euroquis.nl/blabla/2024/07/31/gpg.html )

adriaandegroot:
I'll just do it upstream instead.

PR #401, opened 2024-08-04
handbook/security: fix incorrect command example

rlankfo:
This section indicates to check which password hash is used, one should `grep user /etc/master.passwd`, but then the output of `grep passwd_format /etc/login.conf` is shown.


pauamma (approved):

dbaio:
Committed on https://github.com/freebsd/freebsd-doc/commit/32203a5383329a43bc1b755f522f181298719f19 with a slight change; dropped the command to a normal user.
Thank you!

PR #408, opened 2024-08-27
doc(ZFS): update UEFI loader instructions for 13+

msimerson:
FreeBSD 13.0+ no longer have boot1.efifat, so replace those instructions with the relevant commands and add reference to loader.efi man page.


bsdimp (approved):
This looks good to me.

documentation/content/en/books/handbook/zfs/_index.adoc
```diff
@@ -1335,11 +1335,13 @@ For systems using EFI to boot, execute the following command:
 
 [source,shell]
 ....
-# gpart bootcode -p /boot/boot1.efifat -i 1 ada1
+# mount_msdosfs /dev/ada1p1 /boot/efi
```
jrtc27, line 1338: By default this will automatically be mounted (bsdinstall puts it in fstab) and so will fail with `mount_msdosfs: /dev/ada1p1: Device busy`. If you need to mount it manually for whatever reason, mount -t msdosfs is better than mount_msdosfs (which is a bit of an implementation detail).

documentation/content/en/books/handbook/zfs/_index.adoc
```diff
@@ -1335,11 +1335,13 @@ For systems using EFI to boot, execute the following command:
...
+# cp /boot/loader.efi /boot/efi/efi/boot/bootx64.efi
```
jrtc27, line 1339: This assumes amd64

documentation/content/en/books/handbook/zfs/_index.adoc
msimerson: Correction, it "may fail." You're assuming the disks were created with bsdinstall. There are more than a few FreeBSD disks roaming about that were installed with sysinstall or were manually partitioned to match disks installed long ago.

documentation/content/en/books/handbook/zfs/_index.adoc
msimerson: Fair point. Most of the other FreeBSD docs I've seen *only* copy loader.efi to bootx64.efi, as that's correct likely ~99% of the time. The [wiki page](https://wiki.freebsd.org/UEFI) and [man page](https://man.freebsd.org/cgi/man.cgi?query=uefi) note there is no 32-bit UEFI support, so 'x64' and 'aa64' would cover probably 99.9% of cases.

```sh
# cp /boot/loader.efi /boot/efi/efi/boot/bootx64.efi
or
# cp /boot/loader.efi /boot/efi/efi/boot/bootaa64.efi
```

To cover all cases, perhaps borrow from the convention used in the loader.efi man page?

```sh
# cp /boot/loader.efi /boot/efi/efi/boot/bootXXX.efi (see uefi(8) for values to replace ‘XXX’ with)
```

I don't have enough knowledge to explain why we don't just direct users to *only* copy it to the designated FreeBSD spot:

```sh
# cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi
```

documentation/content/en/books/handbook/zfs/_index.adoc
jrtc27: I realise. I deliberately said *by default* at the start of the sentence to fully acknowledge there are other ways to get a working FreeBSD system. That is, by default it will fail. That doesn't mean it will always fail.

msimerson: How about this?

```sh
mount -t msdosfs -u /dev/ada1p1 /boot/efi
```

sergio-carlavilla:
Any news with this patch?
@jrtc27 @msimerson

documentation/content/en/books/handbook/zfs/_index.adoc
jrtc27: This needs to follow the same approach as loader.efi(8). Both locations are important, but which is more important depends on your system. Only updating the removable media path would be wrong for most desktop and server systems. Only updating the FreeBSD-specific path would be wrong for most SBC systems.

msimerson:
The current instructions are still completely broken for *supported* versions of FreeBSD. 

This patch makes them much less broken, but only for the vast majority of FreeBSD users. My interest in improving the docs has been exhausted by perfect standing in the way of better.

jrtc27:
> The current instructions are still completely broken for _supported_ versions of FreeBSD.
> 
> This patch makes them much less broken, but only for the vast majority of FreeBSD users. My interest in improving the docs has been exhausted by perfect standing in the way of better.

No. The instructions you have provided *do not work for the vast majority of FreeBSD users*, because they only update the removable media path, which is *not* what a normal desktop/server installed via bsdinstall will be using.

grahamperrin:
Open bugs reports include: 

- [255318 – handbook: Document how to update the bootloader](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=255318)
- [263322 – Explain when and why to upgrade the bootloader](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=263322)

<https://github.com/freebsd/freebsd-doc/pull/408#issuecomment-2600971408>

> … My interest in improving the docs has been exhausted …

@msimerson if it helps you to feel better about the difficulties with documenting things such as this: for a (very) long time, I did things wrong because my brain switched off after `efibootmgr -v` in 2023 produced nothing like what's shown under <https://www.freebsd.org/releases/14.0R/relnotes/#upgrade>. 

My own routine includes this, whilst an upgraded but not yet active boot environment is mounted at `/tmp/up`: 

`cp /tmp/up/boot/loader.efi /boot/efi/efi/freebsd/loader.efi && cp /tmp/up/boot/loader.efi /boot/efi/efi/boot/bootx64.efi`

msimerson:
I do a similar thing, but I *always* add /etc/fstab entries with `noauto` for every one of the disks that has boot loaders. Eg: 

```sh
➜ grep efi /etc/fstab
/dev/gpt/nda0-efi	/boot/efi	msdosfs	rw,noauto	2	2
/dev/gpt/nda1-efi	/boot/efi	msdosfs	rw,noauto	2	2
```
I mostly run FreeBSD on servers with mirrored boot disks and those fstab entries are my reminder to update the boot loader on *both* disks. They also make this little saved snippet work:

```sh
df /boot/efi | grep /boot/efi && umount /boot/efi
for _fs in $(grep efi /etc/fstab | awk '{ print $1 }'); do
	mount -t msdosfs -o longnames $_fs /boot/efi
	cp /boot/loader.efi /boot/efi/efi/boot/boot*.efi
	cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi
	umount /boot/efi
done
```

documentation/content/en/books/handbook/zfs/_index.adoc
```diff
@@ -1331,12 +1331,14 @@ For legacy boot using GPT, use the following command:
 # gpart bootcode -b /boot/pmbr -p /boot/gptzfsboot -i 1 ada1
```
grahamperrin, line 1331: `ada1` was assumed in 043d1f2004aee201ba00f65cffaab7e6feb3c31b (2018), maybe better now to assume `ada0` here and below. 

(A separate pull request, to keep this one simple. [261212 – Update the ZFS chapter of the FreeBSD Handbook, and other OpenZFS-related pages](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=261212) is _very_ broad.)

msimerson: as you wish.

documentation/content/en/books/handbook/zfs/_index.adoc
```diff
@@ -1342,18 +1342,22 @@ For legacy boot using GPT, use the following command:
...
 
-For systems using EFI to boot, execute the following command:
+For systems using EFI to boot, execute the following commands:
 
 [source,shell]
 ....
-# gpart bootcode -p /boot/boot1.efi -i 1 ada1
+# umount /boot/efi
+# mount -t msdosfs /dev/ada0p1 /boot/efi
+# cp /boot/loader.efi /boot/efi/efi/boot/boot*.efi
```
bsdimp, line 1354: The '*' here doesn't make sense.
It should be to bootx64.efi.
With a note somewhere saying that's architecture dependent (but ada is primarily a x86 thing) and please refer to this table for the name of other ports. I think that table is in loader.efi, in the examples.

msimerson: My newer devices present at `ndaN` because they're NVMe. Disks attached via SAS/SCSI (the CAM layer) present as `daN`.  Disks attached via (S)ATA present at `adaN`. Some RAID controllers present differently. There's no universal command that'll work there, so perhaps it needs to be: 

```sh
export BOOT_DEVICE="/dev/{daN | ndaN | adaN}"  # CUSTOMIZE THIS!
mount -t msdosfs "$BOOT_DEVICE" /boot/efi
```

The copy command with the `*` works across TIER 1 (and probably all other) supported architectures:

```sh
# ls /boot/efi/efi/boot/boot*.efi   
/boot/efi/efi/boot/bootaa64.efi  # uname -p = aarch64
/boot/efi/efi/boot/bootx64.efi   # uname -p = amd64
```

msimerson: If there was a reasonable chance of it getting merged, I'd update this PR with this:

```sh
# Common EFI: /dev/nda0 (NVMe), /dev/ada0 (SATA), /dev/da0 (USB/SAS/SCSI)
# Note: For mirrored boot drives, update BOTH disks.

# export EFI_PARTITION="/dev/{ndaN | daN | adaN}"  # CUSTOMIZE THIS!
# mount | grep -q /boot/efi && umount /boot/efi
# mount -t msdosfs "$EFI_PARTITION" /boot/efi
# cp /boot/loader.efi /boot/efi/efi/boot/boot*.efi 
# cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi
# umount /boot/efi
```

msimerson: This is more robust, and compared the more static commands above:
- installs when no boot loader exists
- works with all EFI platforms.
  - ARCH matches the two exceptions to the boot{ARCH}.efi naming convention.
  - other architectures (arm) work by default
- doesn't require users to figure out partition names
- improves on my in-house script that relied on an up-to-date `/etc/fstab`
  - EFI requires gpart partitions

```sh
# discover the EFI name for this systems CPU architecture
export ARCH=$(case "$(uname -p)" in aarch64) echo "aa64" ;; amd64) echo "x64" ;; *) echo "$(uname -p)" ;; esac)

for _efi_part in $(gpart show -p | grep ' efi ' | awk '{ print $3 }')
do
  echo; echo "    updating $ARCH EFI boot loader in $_efi_part"

  mount | grep -q 'on /boot/efi ' && umount /boot/efi
  mount -t msdosfs "/dev/$_efi_part" /boot/efi
  mkdir -p /boot/efi/efi/boot /boot/efi/efi/freebsd

  cp /boot/loader.efi "/boot/efi/efi/boot/boot${ARCH}.efi"
  cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi

  umount /boot/efi
done

```

## Sample Output

### amd64 (Intel)

```sh
    updating x64 EFI boot loader in mmcsd0p1
mount | grep -q /boot/efi && umount /boot/efi
mount -t msdosfs /dev/mmcsd0p1 /boot/efi
cp /boot/loader.efi /boot/efi/efi/boot/bootx64.efi
cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi
umount /boot/efi

    updating x64 EFI boot loader in nda0p1
mount | grep -q /boot/efi && umount /boot/efi
mount -t msdosfs /dev/nda0p1 /boot/efi
cp /boot/loader.efi /boot/efi/efi/boot/bootx64.efi
cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi
umount /boot/efi

    updating x64 EFI boot loader in nda1p1
mount | grep -q /boot/efi && umount /boot/efi
mount -t msdosfs /dev/nda1p1 /boot/efi
cp /boot/loader.efi /boot/efi/efi/boot/bootx64.efi
cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi
umount /boot/efi

    updating x64 EFI boot loader in nda2p1
mount | grep -q /boot/efi && umount /boot/efi
mount -t msdosfs /dev/nda2p1 /boot/efi
cp /boot/loader.efi /boot/efi/efi/boot/bootx64.efi
cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi
umount /boot/efi
```

### ARM 64

```sh
    updating aa64 EFI boot loader in nda0p1
mount | grep -q /boot/efi && umount /boot/efi
mount -t msdosfs /dev/nda0p1 /boot/efi
cp /boot/loader.efi /boot/efi/efi/boot/bootaa64.efi
cp /boot/loader.efi /boot/efi/efi/freebsd/loader.efi
umount /boot/efi
```

grahamperrin:
https://github.com/freebsd/freebsd-doc/commit/f8d0c45d7f71a7e5e539188b4662eba41fd9513a

PR #409, opened 2024-09-04
Update Chapter 2: Section 2.5.3 zh-tw translation to the latest version

arnoldsky-yu:
This commit updates the translation of Figure 8 and the accompanying text in Chapter 2.5.3 to the latest English version. The following changes were made:

1. Updated translation of the components selection:
   - The English version has been updated with new components such as `base-dbg`, `kernel-dbg`, `lib32-dbg`, and `tests`, which were not present in the previous Traditional Chinese version. These components are now translated and included.

2. Removed outdated descriptions:
   - The original Traditional Chinese version mentioned `doc` and `games` components, which are no longer listed in the latest English version. These outdated components have been removed to match the English text.


debdrup:
The translation of the documentation is handled through weblate, so I'd recommend checking that out [here](https://docs.freebsd.org/en/books/fdp-primer/weblate/).

PR #410, opened 2024-09-04
sockets(examples): remove unused register keyword, replace bzero with memset

rilysh:
\* Remove the redundant register keyword from the variables, as most compilers (GCC/Clang) just ignore them, and they're mostly just noise.

\* Replace bzero() with memset(). IEEE 1003.1-2008 removes the spec of bzero(), replacing it with memset().

\* Use socklen_t (unsigned type) for variable b in the socket bind example.

\* Add unistd.h header file in the socket connect example (it is used by the write() and read() libc calls). Remove unnecessary (char *) cast from argv[1].

### DIFF
--- a/documentation/content/en/books/developers-handbook/sockets/_index.adoc
+++ b/documentation/content/en/books/developers-handbook/sockets/_index.adoc
@@ -562,8 +562,7 @@ We now know enough to write a very simple client, one that will get current time
 #include <unistd.h>
 
 int main() {
-  register int s;
-  register int bytes;
+  int s, bytes;
   struct sockaddr_in sa;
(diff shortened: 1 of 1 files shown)


debdrup:
I've adjusted the commit message slightly to fit within the columniation that FreeBSD uses, and I've sought review from emaste just to double-check things, so I'm gonna commit this.

Thanks :)

debdrup:
Since I forgot to mention the PR in the commit, I'll just add that it's committed as [920aa82eb14](https://cgit.freebsd.org/doc/commit/?id=920aa82eb14f94dfd90648617bf6cc3df1e68715).

Thanks again.

PR #413, opened 2024-09-17
Update _index.adoc

ajoly18:
Petite erreur corrigée sur le paragraphe en partance de la ligne n°79.

"Soit a machine sur leur bureau." --> "Soit la machine sur leur bureau."


concussious:
Thank you for this submission! I need a real email to take this patch. From the CONTRIBUTING: 

Author Name and Email
Please use a name and email address in the Signed-off-by trailer and as the author of the git commits, since we cannot accept anonymous contributions. It is common, but not required, to use some form of your full name. We realize that some contributors are not comfortable doing so or prefer to contribute under a pseudonym, preferred name, chosen name or similar moniker that might not match relevant government records. We can accept your patch, as long as the name and email you use are distinctive, identifying, and not misleading. Please note that if your patch is accepted, the name and email address will become a permanent and immutable part of the public history of the FreeBSD source tree.

The goal of this policy is to allow us to have sufficient information to contact you if questions arise about your contribution. Addresses of the form something@users.noreply.github.com do not meet this standard since we cannot use it to contact you. We can use .mailmap for name or email changes and mistakes, but is imperfect.

Core (core@freebsd.org) may request authors of significant changes using an obvious pseudonym to document their identity more concretely should there be issues in the future. The author may request this documentation be kept confidential unless needed for legal issues arising from their contributions.

https://github.com/concussious/freebsd-src/blob/main/CONTRIBUTING.md

concussious:
Ping, I would love to merge this but I need the commit to be amended with a real email or I'm in trouble for not doing what I'm supposed to do.

concussious:
Thanks @bsdimp, I will do this now.

PR #427, opened 2024-10-13
handbook: Correct package name in the linuxemu chapter

dzvon:
The correct package name is `emulators/linux-rl9`.


dzvon:
The PR is closed since this has already been fixed in the `main` branch.

PR #430, opened 2024-10-15
porters-handbook/uses(en): add possible arguments for cargo

rilysh:
References:
1. https://doc.rust-lang.org/cargo/commands/cargo-build.html
2. https://doc.rust-lang.org/cargo/commands/cargo-test.html
3. https://doc.rust-lang.org/cargo/commands/cargo-clean.html


dbaio:
Hi,

I think there was some confusion here.

These arguments reference what the ports framework supports for `USES=cargo`. See https://github.com/freebsd/freebsd-ports/blob/main/Mk/Uses/cargo.mk#L4-L6.

For instance, in Python, you can use this type of argument: `USES=python:3.8+	# Supports Python 3.8 or later`

PR #432, opened 2024-10-21
Corrected grep command

ohreally:
The handbook says to grep /etc/master.passwd for the user password hash mechanism, but this should be /etc/login.conf.

### DIFF
--- a/documentation/content/en/books/handbook/security/_index.adoc
+++ b/documentation/content/en/books/handbook/security/_index.adoc
@@ -180,7 +180,7 @@ The following command can be run to check which hash mechanism is currently bein
 
 [source,shell]
 ....
-# grep user /etc/master.passwd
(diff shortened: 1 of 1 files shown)


documentation/content/en/books/handbook/security/_index.adoc
```diff
@@ -180,7 +180,7 @@ The following command can be run to check which hash mechanism is currently bein
 
 [source,shell]
 ....
-# grep user /etc/master.passwd
+# grep passwd_format /etc/login.conf
```
pauamma, line 183: This is consistent with the output given below, but noting that (at least on 13.3)
> % ls -l /etc/login.conf
> -rw-r--r--  1 root  wheel  7094 Jul 11 02:47 /etc/login.conf
so maybe use `%` instead of `#`?

PR #439, opened 2024-12-03
sidenav.html (sidebar): 14.2-RELEASE

grahamperrin:
Fix fdba73cadd55, which omitted a change from rel141 to rel142 in part of one line.


grahamperrin:
@cperciva maybe; or simple enough for anyone to approve. 

Affects: <https://www.freebsd.org/releases/14.2R/> and related pages. 

Thanks

cperciva:
I always screw up something with my last-minute edits.  Sigh.

Thanks for catching this, I've committed the fix!

grahamperrin:
:-) I never think of this type of thing as a true screw-up. Don't sigh (don't be hard on yourself). 

Everybody forgets the sidebar, in the same way that [nobody expects the Spanish Inquisition](https://www.startpage.com/do/dsearch?query=Nobody+expects+the+Spanish+Inquisition&cat=web), and it has taken maybe two years for me to remember that the word is sidenav, not sidebar, for search purposes for FDP purposes for post-release purposes, which is why it put the word _sidebar_ in parentheses in the title of this PR (not to be confused with a PR). 

----

Key phrase, for future recollection without bookmarking:

- _Barbara Stanwyck sideblah_

PR #442, opened 2024-12-08
14.2R/errata: news: inclusion of Wi-Fi firmware

grahamperrin:
Expand upon the second of the four highlights that were announced on 3rd December.

### DIFF
--- a/website/content/en/releases/14.2R/errata.adoc
+++ b/website/content/en/releases/14.2R/errata.adoc
@@ -78,4 +78,5 @@ In particular, instructions for automatically unlocking directories at login tim
 [[late-news]]
 == Late-Breaking News
 
-No late-breaking news.
+* The announcement of 14.2-RELEASE did highlight the installer's ability to download and install firmware packages _after installing_ the FreeBSD base system, but omitted to mention addition of firmware to the images for installers.
+These additions are to automate, or simplify, Wi-Fi connection to the Internet &#8211; during installation &#8211; for the broadest possible range of use cases at this time.


grahamperrin:
Affects: 

- _late-breaking news_ at <https://www.freebsd.org/releases/14.2R/errata/#late-news>. 

<https://old.reddit.com/r/freebsd/comments/1gxn0qf/freebsd_142rc1_now_available/lyi74ou/> (@cperciva, 2024-11-23) noted, and I'll emphasise: 

> The biggest change in 14.2-RC1 is that the **installer images contain some wifi firmware**, and the installer knows how to download and install firmware packages.
> 
> This is important for solving the **chicken-and-egg** problem of systems which need firmware packages in order to connect to the internet – it's **hard to download and install firmware if you don't already have the firmware you need to connect to the internet**. …

<https://www.freebsd.org/releases/14.2R/announce/>, second point: 

> - The installer now supports downloading and installing necessary firmware packages after installing the FreeBSD base system.

It's certainly a highlight, however the phrase lacked the chicken/egg context. 

From [FreeBSD 14.2: What’s new, and how did we get here? | FreeBSD Foundation](https://freebsdfoundation.org/blog/freebsd-14-2-whats-new-and-how-did-we-get-here/): 

> **Installer Improvements**: The bsdinstall(8) utility now supports post-installation firmware downloads​.

– similarly lacking the chicken/egg context. 

----

<https://forums.theregister.com/forum/all/2024/12/05/freebsd_142/#c_4978438> **did** note new capabilities. A valid question remained: 

> So no LAN and only WiFi how am I supposed to 'download' these firmwares …

@jrtc27 offered an answer there (thank you). 

Let's have a comparable statement in an official area. A single, authoritative point of reference. Late-breaking news seems appropriate. 

From the Foundation's post: 

> … Before starting the installation, it is advisable to check the release errata for any recent issues or updates.y recent issues or updates.

(I'll contact Marketing for a correction to that sentence.)

Thanks

concussious:
Overcome by events? When I put my user glasses on, this seems like tl;dr "wifi should work in the installer".

cperciva:
Yes wifi should work in the installer.  The installer images should have all the necessary firmware installed.

PR #461, opened 2025-01-25
Wayland: Add configuration about screen sharing

lbartoletti:
After help from @bapt and @ifreund on IRC, this is a proposal to add Wayland documentation about settings to enable screen sharing.

### DIFF
--- a/documentation/content/en/books/handbook/wayland/_index.adoc
+++ b/documentation/content/en/books/handbook/wayland/_index.adoc
@@ -723,3 +723,129 @@ bar {
 	  command[mouse-left]  =   /usr/local/bin/thunderbird;
 }
 ....
+
+[[wayland-screensharing]]
+== Screen Sharing on Wayland
+
+Screen sharing on Wayland requires additional configuration compared to Xorg due to the differences in how Wayland handles display protocols. To enable screen sharing, you will need to set up tools like `xdg-desktop-portal`, a backend implementation compatible with your compositor, and `pipewire`. These tools work together to provide screen sharing functionality in Wayland environments. The key to making screen sharing work is ensuring that **DBus is properly activated** and that the necessary environment variables, particularly `WAYLAND_DISPLAY`, are correctly set.
+
+=== Required Packages
+
+First, install the necessary packages:
+
+[source,shell]
+----
+# pkg install xdg-desktop-portal pipewire
+----
+
+Next, install the appropriate backend implementation for your compositor. For example:
+
+- **Sway** or other wlroots-based compositors:  
+  [source,shell]
+  ----
+  # pkg install xdg-desktop-portal-wlr
+  ----
+
+- **KDE Plasma**:  
+  [source,shell]
+  ----
+  # pkg install plasma5-xdg-desktop-portal-kde
+  ----
+
+- **GNOME**:  
+  [source,shell]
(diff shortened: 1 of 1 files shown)


dbaio (changes requested):
Hi,

Thank you @lbartoletti  for submitting this.

I have a few points about the overall structure, as it is not rendering well. Please use one sentence per line, following the current style of the chapter. Also, please reduce the usage of the word "you."

[Attached](https://github.com/user-attachments/files/18633741/wayland.patch) is a patch for this PR with my suggestions. Take a look and see if you agree or if you want to change anything else.

Just a note here: the author is a FreeBSD committer and can push this by himself to the doc repository.

ifreund (commented):
Overall this looks a nice improvement. There are a couple of points that I think could use some clarification however:

1. xdg-desktop-portal reads the `XDG_CURRENT_DESKTOP` environment variable to determine which configuration file in the `xdg-desktop-portal` directory to use. For example, setting `XDG_CURRENT_DESKTOP=sway` is necessary to cause it to use the `sway-portals.conf` file in the example given.

2. Running `dbus-update-activation-environment ...` is necessary to ensure that the WAYLAND_DISPLAY (and perhaps XDG_CURRENT_DESKTOP) environment variables are set in the environment of the xdg-desktop-portal dbus service when it is started through dbus activation (e.g. when Firefox accesses the screen sharing dbus interface and xdg-desktop-portal is not yet running).

3. If there is no dbus session bus already running when `dbus-update-activation-environment` is run, `dbus-update-activation-environment` will automatically start one. However this is not necessarily desirable as the automatically started dbus session bus instance is not exposed through the `DBUS_SESSION_BUS_ADDRESS` environment variable and there is a risk of accidentally starting multiple session buses that are not aware of each other such that firefox does not connect to the session bus where `xdg-desktop-portal` is running.

    It would be more robust to explicitly start the dbus session bus before the Wayland compositor is started. Linux distros using systemd or alternatives like [turnstile](https://github.com/chimera-linux/turnstile) start exactly one dbus session bus per user that is shared across all login sessions.

    Since we do not currently have infrastructure to manage the dbus session bus in that way on FreeBSD, I think the most robust way to handle this is to set things up in a wrapper script that starts the Wayland compositor:

```sh
#!/bin/sh
export XDG_CURRENT_DESKTOP=sway
# Starts a session bus instance of dbus-daemon(1) and sets $DBUS_SESSION_BUS_ADDRESS
# in the environment before executing sway
exec dbus-run-session sway
```

And then in the running Wayland compositor inform the dbus session bus of the WAYLAND_DISPLAY variable set by the compositor:
```
dbus-update-activation-environment WAYLAND_DISPLAY
```

TheRealKraytonian:
Does this still need to be edited I will do this as this config is crucial.

lbartoletti:
> Does this still need to be edited I will do this as this config is crucial.

Yes my first PR has to be edited with @ifreund comments. Tested and it works well. If you have time, you could supersed my PR.

TheRealKraytonian:
I was unable to get this working? Let me know if you need more info

lbartoletti:
> I was unable to get this working? Let me know if you need more info

What is the problem for you?

I don't work anymore for me too, but because of (nvidia) drm-kmod...

PR #462, opened 2025-02-12
handbook: Remove mount_procfs(8) from basics

spmzt:
mount_procfs(8) no longer exists.


concussious (approved):
For this type of commit, I always mention in the commit message that the underlying thing was removed in commit 9fd5a1b14f34126f68ede5d8a5a2c3a864f0c4e4.

PR #467, opened 2025-02-20
Update press.toml

adickison:
added two new items from the Foundation blog


Jehops:
Hi Anne.  Everything looks good.  I made the minor tweaks we discussed privately, so it's ready to be committed.  I'll do that later today.

adickison:
Committed

PR #471, opened 2025-03-11
releases.adoc :beta2-upcoming: IGNORE

grahamperrin:
```text
13.5-RELEASE is announced, so sidenav.html should cease showing 
13.5 as an upcoming release at pages such as
https://www.freebsd.org/releases/13.5R/

Complements: doc a587e34fca7d for
13.5: Release-related website updates
```

### DIFF
--- a/shared/releases.adoc
+++ b/shared/releases.adoc
@@ -52,7 +52,7 @@
 
 // Set to 'INCLUDE' when the schedule is available.
 
-:beta2-upcoming: INCLUDE
+:beta2-upcoming: IGNORE
 
 // Set to 'INCLUDE' when the installation images are available.
 


grahamperrin:
@cperciva

Whilst I can't quite make sense of how this change dovetails with the sidebar code, I **assume** that:

-  `beta2` (2) will mean any prerelease of 13 for as long as there is one, and only one, superior major branch/version with at least one release (i.e. we have 14.1-RELEASE and 14.2-RELEASE but not yet 15.0-RELEASE)

– or words to that effect. Correct me, please, if I'm wrong. 

Related – in `sidenav.html` for the sidebar: 

https://github.com/freebsd/freebsd-doc/blob/29acf77c151ce7713883f190e5a05a7b12ee6a1f/website/themes/beastie/layouts/partials/sidenav.html#L66-L80

cperciva:
I don't fully understand how the code works, but my understanding is that beta-* and beta2-* are equivalent, you just need to make sure you don't get them mixed up (like I did).  We had 13.5 and 14.3 both marked as upcoming, and I accidentally marked 14.3 as no longer upcoming rather than marking 13.5 as no longer upcoming.

I think it's fixed now...

grahamperrin:
Thanks, I didn't realise that 14.3 should be publicly linked until after I noticed it at <https://www.freebsd.org/> … I guess, it wasn't there until after your fix: 

- 29c5bae51b0bbc868903b1f10089222692ac24cc

PR #472, opened 2025-03-12
Update _index.adoc

MrAriaNet:
added iran mirror official


concussious:
You commit message would need to be something like:
```
handbook/mirrors: Add Iranian mirror
```
For more information, please see:
https://docs.freebsd.org/en/articles/committers-guide/#commit-log-message

concussious:
Hey, you can just `git commit --amend` and it will just update.

PR #475, opened 2025-03-19
status, 2024 q4: sponsors

grahamperrin:
Fix markup for sponsorship lines.

Whilst here, remove superfluous markup.

Consistency: one line per sponsor.

Consistency: The FreeBSD Foundation is a sponsor (not sponsors).


lsalvadore:
I have merged only the part regarding the ends of line. I think it is fine having multiple sponsors on the same line or using 'Sponsors:' with a one item long list of sponsors. We do not have a standard about this in status reports at the moment and I do not see a reason to introduce one, at least for now.

Thanks!

grahamperrin:
Thanks. 

[`f0d0133` (#475)](https://github.com/freebsd/freebsd-doc/pull/475/commits/f0d0133c54bfe52c64321c3f233e1cf24ffff5ae) was because it's singular in fifteen places: 

> Sponsor: The FreeBSD Foundation

<https://www.freebsd.org/status/report-2024-10-2024-12/#_cluster_administration_team> plural is the odd one out.

PR #476, opened 2025-03-19
status, 2024 q4: fix broken links

grahamperrin:
https://www.freebsd.org/status/report-2024-10-2024-12/#_continuous_integration

The originally published link for work in progress presents 123
reports, most of which are not WIP.

Fix the breakage. Reference: example 12 at
https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#links

```text
link:++https://example.org/?q=[a b]++[URL with special characters]

https://example.org/?q=%5Ba%20b%5D[URL with special characters]
```

Starting with `++` instead of `https:` is **exceptional**; 

> … The `link:` macro prefix is not required when the target starts with a URL scheme like `https:`. …

### DIFF
--- a/website/content/en/status/report-2024-10-2024-12/ci.adoc
+++ b/website/content/en/status/report-2024-10-2024-12/ci.adoc
@@ -6,7 +6,7 @@ link:https://tinderbox.freebsd.org[FreeBSD CI Tinderbox view] URL: link:https://
 link:https://artifact.ci.FreeBSD.org[FreeBSD CI artifact archive] URL: link:https://artifact.ci.FreeBSD.org[] +
 link:https://wiki.FreeBSD.org/HostedCI[Hosted CI wiki] URL: link:https://wiki.FreeBSD.org/HostedCI[] +
 link:https://wiki.FreeBSD.org/3rdPartySoftwareCI[3rd Party Software CI] URL: link:https://wiki.FreeBSD.org/3rdPartySoftwareCI[] +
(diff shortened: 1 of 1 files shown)


grahamperrin:
Tested: 

```text
git -C /usr/doc fetch
git -C /usr/doc switch patch-13
grahamperrin:~ % cd /usr/doc/website
grahamperrin:/usr/doc/website % make run DOC_LANG="en" 
```

image

… etc. The three links now present **28** reports. 

----

Local housekeeping (revert the switch): 

```text
grahamperrin:~ % git -C /usr/doc switch -
Switched to branch 'main'
Your branch is ahead of 'origin/main' by 463 commits.
  (use "git push" to publish your local commits)
grahamperrin:~ % git -C /usr/doc pull --ff-only freebsd main && git -C /usr/ports pull --ff-only freebsd main && git -C /usr/src pull --ff-only freebsd main
From https://git.freebsd.org/doc
 * branch                  main       -> FETCH_HEAD
Already up to date.
From https://git.freebsd.org/ports
 * branch                      main       -> FETCH_HEAD
Already up to date.
From https://git.freebsd.org/src
 * branch                      main       -> FETCH_HEAD
Already up to date.
grahamperrin:~ % 
```

----

I guess, plain text versions of future reports (for email etc) might require special attention where URLs are marked up in this way. 

Hint: if you see italics in a link, check for breakage. 

image

lsalvadore:
Merged, thanks!

grahamperrin:
Thanks! Re: the commit log message, 

> Use of special characters such as '_' in links …

– it's not exactly a special character, and not link-specific.

Better to think of what's in this commit as uncommon use of pairs of a normal ASCII character, which might occur anywhere (not necessarily in a link). 

I make this clarification because (please) `++` should not be overused. Same as overuse of the `link:` prefix, it makes proofreading difficult.

<https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#text-formatting>, the _un_constrained italic example: 

image

PR #479, opened 2025-03-26
Add openjdk status report for 2025Q1

snake66:
Here's my status report for the Improve OpenJDK on FreeBSD project.

Sorry it's a bit late, but I wanted to get the jdk24 port mostly in place before submitting it.


lsalvadore (approved):
Approved once a few small nits are fixed and the report is switched to one-sentence-per-line style. I can do these at merge time if your prefer.

lsalvadore:
Merged, thanks!

PR #491, opened 2025-04-22
doc: network: Add default ctrl_interface and related parameters to wpa_supplicant.conf

TheRealKraytonian:
This commit adds the following default configuration parameters to the example wpa_supplicant.conf in the handbook:

ctrl_interface=/var/run/wpa_supplicant
eapol_version=1
ap_scan=1
fast_reauth=1

This is necessary for users to effectively disable, debug, and manage their wireless network connections. These parameters are already present in the example configuration files provided with wpa_supplicant, suggesting they are commonly used and beneficial for a standard setup. Including them in the handbook's basic example will provide a more robust and user-friendly default configuration for wireless networking.

PR: 286187
Reported by: kraytonian hello@chingarande.com
Reviewed by: ziaee ziaee@freebsd.org
Tested by: kraytonian hello@chingarande.com
Discussed with: ziaee@freebsd.org


concussious:
A better commit message might be:
```
Handbook/network: Mention wpa_supplicant.conf boilerplate

Mention wpa_supplicant.conf boilerplate required for the wpa_cli wireless network manager. These parameters are already present in the share/examples/wpa_supplicant.conf.

PR: 286187
Reported by: <realnameplease> hello@chingarande.com
Discussed with: ziaee@freebsd.org
```
Key points:
- handbook/network shows what is actually changing
- keep commit message titles terse for one line, many git tools limit to 50 characters
- no need to mention what, they can see that, the commit message needs to explain why
- check that path, I'm on my phone at work
- never say you tested something you authored, this is expected and required
- discussed with xor reviewed by, in this case, we talked about it but I hadn't seen a diff until just now
- you need to use your real name when changing global critical infrastructure

concussious (approved):

PR #497, opened 2025-05-10
Translated arch-handbook/vm/chapter.adoc to Russian

VladlenPopolitov:
1. Added arch-handbook/vm/chapter.adoc (Virtual Memory Management in FreeBSD Architecture Handbook) and translated to Russian.
2. Added the abstract into arch-handbook/book.adoc, translated to Russian from English version


VladlenPopolitov:
I found correct way to commit translation in weblate.

PR #499, opened 2025-05-12
Escape a + sign in porters handbook/makefiles

snake66:
The first plus sign in the "Detailed Information" section is interpreted as the start of a literal monospace section, and breaks formatting. This may be a bug in AsciiDoctor, but escaping it avoids the issue.


sergio-carlavilla:
Fixed: https://cgit.freebsd.org/doc/commit/?id=41d8ee28f24de8a4442aee3018fab1e42fbc8a39
Thanks!

PR #507, opened 2025-06-14
[KDE][Doc] Add `#` before root commands

osalbahr:


sergio-carlavilla:
Fixed: https://cgit.freebsd.org/doc/commit/?id=27d743b2a71df6917bb549b28c4a85928562b864
Thanks!

PR #512, opened 2025-06-21
Remove gbde from Handbook

ykla:
GBDE is deprecated and will be removed in 15.0, See also <https://lists.freebsd.org/archives/freebsd-current/2023-February/003300.html>, <https://github.com/freebsd/freebsd-src/commit/8d2d1d651678178aa7f24f0530347f860423fd9e>


concussious:
Thanks ykla, but because it's removed in 15, we can't remove it until 2028 when the last release containing it is no longer supported.

It would be nice to instead mention that it's deprecated and is not present in FreeBSD 15.

PR #515, opened 2025-06-25
Add OpenJDK status report for 2025Q2

snake66:
Here's my 2025Q2 status report for the Improve OpenJDK on FreeBSD project.


lsalvadore:
Merged, thanks!

PR #522, opened 2025-07-08
handbook: fix missing space after period in 4.2

troll5:
Fixed missing space after period in 4.2.

Before: "approach.In some cases"
After: "approach. In some cases"


sergio-carlavilla:
Fixed in: https://cgit.freebsd.org/doc/commit/?id=18edf87bc3e7d2281494f8363082cb2fd8687092
Thanks!

PR #524, opened 2025-07-15
Update consulting.toml

phips:
Remove dead site (404)


sergio-carlavilla:
Done in: https://cgit.freebsd.org/doc/commit/?id=f009d148153400e57d9e0bac961ce80743880ee1
Thanks!

PR #525, opened 2025-07-15
Update consulting.toml

FlorineFreeBSD:
Removing psychsoftek " dead site"


Jehops:
Thanks.  Committed with tweaks to the commit message.

The FreeBSD Committer's Guide has some information about commit log messages:
https://docs.freebsd.org/en/articles/committers-guide/#commit-log-message

PR #528, opened 2025-07-18
handbook/x11/nvidia: Fix matching quotation

osalbahr:


osalbahr:
Sure! It's osalbahr@gmail.com

sergio-carlavilla:
Fixed in: https://cgit.freebsd.org/doc/commit/?id=c49bd0d688f8f141ab90922976f7026df032b4b6
Thanks!

PR #529, opened 2025-07-18
handbook/cutting-edge: Add `pkg update` and `pkg upgrade`

osalbahr:
The page title is "Chapter 26. Updating and Upgrading FreeBSD". Which I think should include upgrading packages. `pkg upgrade` is mentioned in  passing but I think it should be explicitly mentioned along with  `pkg update`.

### DIFF
--- a/documentation/content/en/books/handbook/cutting-edge/_index.adoc
+++ b/documentation/content/en/books/handbook/cutting-edge/_index.adoc
@@ -60,6 +60,7 @@ This chapter describes how to track the development system and the basic tools f
 
 After reading this chapter, you will know:
 
+* How to keep a FreeBSD system packages up-to-date with with `pkg update` and `pkg upgrade`.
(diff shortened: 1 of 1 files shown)


pauamma:
Perhaps instead/in addition:
- since pkgbase is in 15-CURRENT, perhaps mention using that to upgrade the base OS?
- link to https://docs.freebsd.org/en/books/handbook/ports/#pkgng-upgrading

What do others think?

grahamperrin:
FYI: 

1. <https://reviews.freebsd.org/D51279> (closed)
2. <https://reviews.freebsd.org/D51279#1190907> "… we can improve it later in small commits. …"
3. https://github.com/freebsd/freebsd-doc/commit/fb42d9b2fb16449812031a307f1b35a047a97f9c (the outcome of D51279)
4. https://github.com/freebsd/freebsd-doc/pull/542

PR #533, opened 2025-07-26
status: fix links

grahamperrin:
Fix broken URLs in four status reports and a template.

### DIFF
--- a/website/content/en/status/report-2024-01-2024-03/doceng.adoc
+++ b/website/content/en/status/report-2024-01-2024-03/doceng.adoc
@@ -23,8 +23,8 @@ mailto:trhodes@FreeBSD.org[Tom Rhodes]'s doc commit bit was taken for safekeepin
 
 ==== FreeBSD Translations on Weblate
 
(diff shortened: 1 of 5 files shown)


lsalvadore:
Merged, thanks!
I have refrained from changing the status template (at least for now) because this is the only instance of this mistake.

@fernape: I think doceng uses its own template to write reports. Please make sure to fix this issue in your template so that the mistake is not repeated in the future. Thanks!

fernape:
El mar, 5 ago 2025, 11:53, Lorenzo Salvadore ***@***.***>
escribió:

> *lsalvadore* left a comment (freebsd/freebsd-doc#533)
> <https://github.com/freebsd/freebsd-doc/pull/533#issuecomment-3154435964>
>
> Merged, thanks!
> I have refrained from changing the status template (at least for now)
> because this is the only instance of this mistake.
>
> @fernape <https://github.com/fernape>: I think doceng uses its own
> template to write reports. Please make sure to fix this issue in your
> template so that the mistake is not repeated in the future. Thanks!
>
Hello

I'm not the doceng secretary anymore.

Cheers

PR #534, opened 2025-07-26
status: releng: fix links

grahamperrin:
13.5/announce/
should be
13.5R/announce/

14.3/announce/
should be
14.3R/announce/


grahamperrin:
@cperciva

cperciva:
Sigh.  I dunno how I did that.  I guess a copy-and-paste error...

Thanks for catching it, patches committed!

PR #536, opened 2025-07-28
documentation: Correct link to nginx-lite

phips:
I hope :) Clicking link as is doesn't work, maybe full description in ports, like this, will? I've not been able to test this via GH edit link. Please check.


sergio-carlavilla:
Thanks, I'll commit it

sergio-carlavilla:
Fixed in: https://cgit.freebsd.org/doc/commit/?id=59c2ec09f72975d3883f19d657c1bf207694a810

PR #557, opened 2025-10-01
Sync sshd policy example with freebsd-src.

skhal:
The default ``sshd'' policy from freebsd/freebsd-src is available at: https://github.com/freebsd/freebsd-src/blob/e7258c42520c18c034f644b97377d8d2a0ad8b7c/lib/libpam/pam.d/sshd

This change updates the default policy example with the source code.


skhal:
closing due to inactivity

PR #594, opened 2026-01-07
In "Latest Ports" section: FreeBSD --> FreeBSD-ports

briandorsey:
It looks like the name has changed as of 15-RELEASE. Found the fix via this forum thread: https://forums.freebsd.org/threads/error-when-switching-to-latest-pkg-repo.100579/

NOTE: That thread also recommends moving `FreeBSD-ports-kmods` to latest as well by adding:
```
FreeBSD-ports-kmods: {
  url: "pkg+https://pkg.FreeBSD.org/${ABI}/kmods_latest_${VERSION_MINOR}"
}
```

Should that be added to this section as well?

### DIFF
--- a/documentation/content/en/books/handbook/ports/_index.adoc
+++ b/documentation/content/en/books/handbook/ports/_index.adoc
@@ -201,7 +201,7 @@ To switch man:pkg[8] from Quarterly to Latest run the following commands:
 [source,shell]
 ....
 # mkdir -p /usr/local/etc/pkg/repos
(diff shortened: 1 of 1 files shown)


concussious (approved):

dag-erling:
This change should not have been merged as-is. The repository has been renamed in FreeBSD 15, but not in FreeBSD 13 or 14. Furthermore, the entire section should be rewritten to teach the reader how repository definitions (and especially partial redefinitions) work, not just give them a command to blindly run without understanding.

PR #614, opened 2026-02-24
Update VirtualBox Guest Section

jlenthe:
Updates are:
* Clarify guest additions package versions
* Use sysrc for rc.conf examples
* Add VM setup section
* Add sub-section headings

It is important to clarify the package versioning since the current handbook calls for installing `virtualbox-ose-additions` which is for VirtualBox 6.1 and shouldn't be used with VirtualBox 7. There are version specific package names that should be used for 7.

Also, it is important to call out `VBoxSVGA` for the graphics controller since `VMSVGA` is the default for FreeBSD guests in VirtualBox 7 and that does not work as well as `VBoxSVGA`. In particular, window scaling doesn't work at all and `VMSVGA` requires a separately installed driver.

I tested Xorg with window scaling and clipboard sharing on FreeBSD 15.0, 14.3, and 13.5 using VirtualBox 7.2.6 on a Windows host.

Thanks for reviewing!


documentation/content/en/books/handbook/virtualization/_index.adoc
```diff
@@ -296,42 +296,85 @@ The virtualization software is available for most common operating systems, incl
...
+[[virtualization-virtualbox-guest-additions]]
+=== Guest Additions Setup
+
+First, install the package:emulators/virtualbox-ose-additions-72[virtualbox-ose-additions] package that corresponds to the version of VirtualBox(TM) installed on the host machine and whether support for X11 is desired or not, as in the following table.
+
+[cols="1,2,2", options="header"]
+|===
+|VirtualBox(TM) Version | Guest Additions Package (with X11) |Guest Additions Package (without X11)
+| 7.2 | package:emulators/virtualbox-ose-additions-72[virtualbox-ose-additions-72] | package:emulators/virtualbox-ose-additions-nox11-72[virtualbox-ose-additions-nox11-72]
+| 7.1 | package:emulators/virtualbox-ose-additions-71[virtualbox-ose-additions-71] | package:emulators/virtualbox-ose-additions-nox11-71[virtualbox-ose-additions-nox11-71]
```
ykla: It lists some EOL version and it will be deleted. See also <https://www.freshports.org/emulators/virtualbox-ose-70/> and <https://www.freshports.org/emulators/virtualbox-ose-71> or <https://www.freshports.org/emulators/virtualbox-ose>

ykla (commented):
Thanks, but I don't think it's necessary to list these tables, as they will be removed soon and are already EOL upstream.

documentation/content/en/books/handbook/virtualization/_index.adoc
jlenthe: Good point. I have removed the table and simplified the guidance in 3b41c2c. Thanks!

PR #630, opened 2026-03-26
handbook: update outdated version references

hypery11:
Couple of things I noticed that are a bit out of date:

- The ntpdate entry in bsdinstall still says it "will soon be retired" — it's been deprecated for a while now, so I reworded it to just say that.
- The STABLE branch explanation in cutting-edge used FreeBSD 10.1 as the example, bumped it to 14.1.

### DIFF
--- a/documentation/content/en/books/handbook/cutting-edge/_index.adoc
+++ b/documentation/content/en/books/handbook/cutting-edge/_index.adoc
@@ -818,7 +818,7 @@ A new checkout of the source is required.
 |`stable/_X_`
 |
 
-The Release version plus all additional development on that branch. _STABLE_ refers to the Applications Binary Interface (ABI) not changing, so software compiled for earlier versions still runs. For example, software compiled to run on FreeBSD 10.1 will still run on FreeBSD 10-STABLE compiled later.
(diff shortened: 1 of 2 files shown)


concussious (approved):
Thanks Po Han!

documentation/content/en/books/handbook/cutting-edge/_index.adoc
```diff
@@ -818,7 +818,7 @@ A new checkout of the source is required.
 |`stable/_X_`
 |
 
-The Release version plus all additional development on that branch. _STABLE_ refers to the Applications Binary Interface (ABI) not changing, so software compiled for earlier versions still runs. For example, software compiled to run on FreeBSD 10.1 will still run on FreeBSD 10-STABLE compiled later.
+The Release version plus all additional development on that branch. _STABLE_ refers to the Applications Binary Interface (ABI) not changing, so software compiled for earlier versions still runs. For example, software compiled to run on FreeBSD 14.1 will still run on FreeBSD 14-STABLE compiled later.
```
concussious: I have an idea to make this even better so we don't have to keep doing this. Give me a week.

concussious: Sorry it took so long, here's the idea so we don't have to do this anymore:

```suggestion
The Release version plus all additional development on that branch. _STABLE_ refers to the Applications Binary Interface (ABI) not changing, so software compiled for earlier versions still runs. For example, software compiled to run on FreeBSD {rel-legacy} will still run on FreeBSD {rel-legacy-major}-STABLE compiled later.
```

hypery11:
Applied, thanks for the suggestion.

PR #633, opened 2026-03-27
website/features: fix hast to hastd under Capsicum

kovan:
The Capsicum section on the features page lists "hast" but there is no hast(8) manual page. The correct name is hastd(8).

PR: 293950


sergio-carlavilla:
Done: https://cgit.freebsd.org/doc/commit/?id=6dae78c5e77082416322b1f921a5a7d50cb9a801

PR #635, opened 2026-03-27
faq: fix single-user mode boot instructions

kovan:
The FAQ referenced a `Boot:` prompt which no longer appears in modern FreeBSD. The boot loader now presents a menu where users can select "Boot Single User" directly, or press Escape to reach the loader prompt.

Changes:
- Update both single-user boot instructions (forgot root password, read-only rc.conf) to describe the boot loader menu
- Simplify `man:exit[1]` to just `exit`

PR: 291048

### DIFF
--- a/documentation/content/en/books/faq/_index.adoc
+++ b/documentation/content/en/books/faq/_index.adoc
@@ -436,10 +436,10 @@ bindkey ^[[3~ delete-char # for xterm
 === I have forgotten the root password! What do I do?
 
 Do not panic!
-Restart the system, type `boot -s` at the `Boot:` prompt to enter single-user mode.
+Restart the system and select `Boot Single User` from the boot loader menu, or press `Escape` at the menu and type `boot -s` at the loader prompt, to enter single-user mode.
 At the question about the shell to use, hit kbd:[Enter] which will display a # prompt.
 Enter `mount -urw /` to remount the root file system read/write, then run `mount -a` to remount all the file systems.
-Run `passwd root` to change the `root` password then run man:exit[1] to continue booting.
+Run `passwd root` to change the `root` password then run `exit` to continue booting.
 
 [TIP]
 ====
@@ -458,7 +458,7 @@ For more information see the section about encrypted disks in the FreeBSD extref
 [[rcconf-readonly]]
 === I made a mistake in rc.conf, or another startup file, and now I cannot edit it because the file system is read-only. What should I do?
 
-Restart the system using `boot -s` at the loader prompt to enter single-user mode.
+Restart the system and select `Boot Single User` from the boot loader menu, or press `Escape` at the menu and type `boot -s` at the loader prompt, to enter single-user mode.
 When prompted for a shell pathname, press kbd:[Enter] and run `mount -urw /` to re-mount the root file system in read/write mode.
 You may also need to run `mount -a -t ufs` to mount the file system where your favorite editor is defined.
 If that editor is on a network file system, either configure the network manually before mounting the network file systems, or use an editor which resides on a local file system, such as man:ed[1].


BigSneakyDuck:
Thanks for this, I submitted the original Bugzilla PR but didn't feel confident enough to submit a patch (or even to give a "proper" review of this one). But I'll flag up two points from the Bugzilla comments for consideration. Firstly https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291048#c4 

> Is there a reason to recommend typing `passwd root` rather than just `passwd`? Specifying user is optional here, see passwd(1):  https://man.freebsd.org/cgi/man.cgi?query=passwd&sektion=1

Also as Graham Perrin points out, booting to single user mode seems (can anyone confirm?) to leave vt using the default USA key map. This may be surprising to e.g. someone using an AZERTY keyboard. This can still be problematic for users with a keyboard similar yet not identical to USA, for instance a standard UK keyboard where the letters will work the same but many special characters do not: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291048#c3

> But this all results in an insidious problem - typing these commands gives no indication of a different keymap, and passwords are not echoed, so anyone who includes an affected special character as part of their new root password may be mystified that it doesn't work once they return to multi-user mode, even though they press exactly the same keys!!

I think the keyboard issue deserves documenting.

bsdimp:
Yes. The keyboard keymap is usually set by the rc scripts, which haven't run yet (obviously) when you're just starting a shell...

BigSneakyDuck:
Another thing that crossed my mind: the FAQ has been almost completely stripped of "how-to" technical details (compare the state of the FAQ 10 years ago) because it's a maintenance burden that can easily become stale. And in fact that's what happened here, in one of the very few places such technical details were left in.

There's surely a strong argument to incorporate this material into the Handbook instead, which I think lacks such explicit instructions, perhaps at:

https://docs.freebsd.org/en/books/handbook/basics/#consoles-singleuser

Then reduce the FAQ answer to a stub, advising that it may be possible to reset the root password by booting into single user mode and following the instructions in the linked Handbook section. A stub plus section link should be more maintainable, so long as content about root password reset is not then moved somewhere else within the Handbook without updating the FAQ. I don't know if FreeBSD documentation guards against this using tracking comments or similar tricks to 

https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Linking#RENAMESECTION

sergio-carlavilla:
> Another thing that crossed my mind: the FAQ has been almost completely stripped of "how-to" technical details (compare the state of the FAQ 10 years ago) because it's a maintenance burden that can easily become stale. And in fact that's what happened here, in one of the very few places such technical details were left in.
> 
> There's surely a strong argument to incorporate this material into the Handbook instead, which I think lacks such explicit instructions, perhaps at:
> 
> https://docs.freebsd.org/en/books/handbook/basics/#consoles-singleuser
> 
> Then reduce the FAQ answer to a stub, advising that it may be possible to reset the root password by booting into single user mode and following the instructions in the linked Handbook section. A stub plus section link should be more maintainable, so long as content about root password reset is not then moved somewhere else within the Handbook without updating the FAQ. I don't know if FreeBSD documentation guards against this using tracking comments or similar tricks to
> 
> https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Linking#RENAMESECTION

This is totally true, and it's "my fault", as you said, we made this to avoid look "outdated".
What do you miss in the FAQ?

BigSneakyDuck:
> What do you miss in the FAQ?

Personally I'm glad the FAQ was stripped down. I know it received a lot of work relatively recently so don't want to rock the boat too much, but I would say:

1. Like many internet FAQs, it can lean more towards "questions we would like to take this opportunity to address" than "questions which other people frequently ask". It might benefit from more user research into popular FreeBSD search queries, questions asked in forums, etc. I complained elsewhere online (and am learning my lesson that it's better to just submit a PR than vent into the ether!) about the tonal shift of the "Funnies", a relic of more free-wheeling days, but I see they've gone now - I thank the authors for making the FAQ less self-indulgent and better focused on new users.
2. Unfortunately there's sometimes no good resource to direct people towards. A common Reddit post is "I'm thinking of switching to FreeBSD, is my graphics card supported?" but this isn't listed in hardware compatibility relnotes - the FAQ attempts to answer this with FreeBSD Wiki links, but one is deleted and the other years out of date. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=290905

I don't know how much traffic the FAQ gets and hence what effort it is worth; probably the most important thing is that its answers are accurate so search engine snippets and LLM training are not misled.

Now Warner has confirmed what I thought about key maps, I'm confident enough to try making a PR which adds explicit root password reset instructions to the Handbook (including a warning about key maps) and reduces the FAQ entry to stub answer plus section link. Let me know if that would be welcome.

sergio-carlavilla:
> > What do you miss in the FAQ?
> 
> Personally I'm glad the FAQ was stripped down. I know it received a lot of work relatively recently so don't want to rock the boat too much, but I would say:
> 
>     1. Like many internet FAQs, it can lean more towards "questions we would like to take this opportunity to address" than "questions which other people frequently ask". It might benefit from more user research into popular FreeBSD search queries, questions asked in forums, etc. I complained elsewhere online (and am learning my lesson that it's better to just submit a PR than vent into the ether!) about the tonal shift of the "Funnies", a relic of more free-wheeling days, but I see they've gone now - I thank the authors for making the FAQ less self-indulgent and better focused on new users.
> 
>     2. Unfortunately there's sometimes no good resource to direct people towards. A common Reddit post is "I'm thinking of switching to FreeBSD, is my graphics card supported?" but this isn't listed in hardware compatibility relnotes - the FAQ attempts to answer this with FreeBSD Wiki links, but one is deleted and the other years out of date. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=290905
> 
> 
> I don't know how much traffic the FAQ gets and hence what effort it is worth; probably the most important thing is that its answers are accurate so search engine snippets and LLM training are not misled.
> 
> Now Warner has confirmed what I thought about key maps, I'm confident enough to try making a PR which adds explicit root password reset instructions to the Handbook (including a warning about key maps) and reduces the FAQ entry to stub answer plus section link. Let me know if that would be welcome.

I think there's no problem to put in both.
And the FAQ gets a lof ot traffic :) also the Handbook.

About this ticket, is merged, I am going to close this PR

BigSneakyDuck:
@sergio-carlavilla 

In the commit I seem to be listed as the author, but I'm just the reporter! https://github.com/freebsd/freebsd-doc/commit/67232f1a2d5e0254a807bebb5fb90d47ff3545b7

The author was kovan.

sergio-carlavilla:
Sorry, my bad

PR #636, opened 2026-03-27
Update _index.adoc

onetruemike:
Corrected nvidia BusID in 20-drivers.conf to match the output of the previous pciconf command.


sergio-carlavilla:
Done: https://cgit.freebsd.org/doc/commit/?id=c085ee0716be0905e6872775f0d64cbe7bdf4855

Thanks! Really good catch!

PR #643, opened 2026-04-05
status/report: add 2026Q1 clusteradm report

ppaeps:
Mostly business as usual this quarter.

Note the cluster refresh for the upcoming stable/13 EoL.  Give a tally of the current machines we have in the cluster and note what the update cadence is.

The FreeBSD Foundation does not fund this work.


lsalvadore:
Merged, thanks!

PR #644, opened 2026-04-05
handbook/bsdinstall: Remove defunct hardening options

BigSneakyDuck:
The bsdinstall hardening script no longer offers knobs for disable_sendmail or enable_aslr.

By default, sendmail is now disabled and ASLR enabled. No need to document this in chapter for fresh installations.

Note: I've checked this applies to all supported releases except 13.5R, which reaches EOL in on April 30, 2026 and is deemed "legacy" by RE so is not recommended for fresh installations anyway. I'm fine if it's preferred not to commit these changes for 3 weeks just in case!


concussious (approved):

concussious:
Thanks for sending this in.

sergio-carlavilla:
Done here: https://cgit.freebsd.org/doc/commit/?id=8970ac5ea149d1ae4e9b99d5974b5d3947c2af19
I think nobody will complain about this, it's only two weeks :D

PR #646, opened 2026-04-06
handbook: fix outdated references and minor grammar issues

hypery11:
Went through a few more handbook chapters and found some stuff that's out of date or just wrong:

- **config**: inetd is referenced as a current way to start services, but it's been deprecated. Reworded to note that.
- **disks/gbde**: the whole gbde example still uses the old ad* device naming (pre-CAM ATA driver). Updated to ada* to match what people actually see on modern FreeBSD.
- **disks/gbde**: there was a NOTE warning about sysinstall being incompatible with gbde. sysinstall has been gone since FreeBSD 10, so I removed it.
- **desktop**: "a easy-to-use" → "an easy-to-use", "different kind of documents" → "different kinds"

### DIFF
--- a/documentation/content/en/books/handbook/config/_index.adoc
+++ b/documentation/content/en/books/handbook/config/_index.adoc
@@ -487,7 +487,8 @@ Additional information can be found in man:rc[8] and man:rc.subr[8].
 
 === Using Services to Start Services
 
-Historically, other services could be started using man:inetd[8], though this has been deprecated in recent FreeBSD releases.
+Other services can be started using man:inetd[8].
+Working with man:inetd[8] and its configuration is described in depth in crossref:network-servers[network-inetd,"The inetd Super-Server"].
 
 In some cases, it may make more sense to use man:cron[8] to start system services.
 This approach has a number of advantages as man:cron[8] runs these processes as the owner of the man:crontab[5].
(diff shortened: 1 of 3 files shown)


documentation/content/en/books/handbook/config/_index.adoc
```diff
@@ -487,8 +487,7 @@ Additional information can be found in man:rc[8] and man:rc.subr[8].
 
 === Using Services to Start Services
 
-Other services can be started using man:inetd[8].
-Working with man:inetd[8] and its configuration is described in depth in crossref:network-servers[network-inetd,“The inetd Super-Server”].
+Historically, other services could be started using man:inetd[8], though this has been deprecated in recent FreeBSD releases.
```
concussious: I think some services are still operated via inetd, like tftpd.

hypery11:
Good call on the inetd thing — you're right, tftpd and others still use it. I reverted that change, the original wording is fine.

concussious (approved):

sergio-carlavilla:
What's left here? Can I proceed?

hypery11:
Nothing left from my side, please go ahead. Thanks!

concussious:
Sorry I didn't get around to this sooner.
I did not touch the gbde section because that is historical and will be removed soon.
Thanks for the patch!

PR #660, opened 2026-04-26
usergroups: Update Portland, remove 5 defunct

BigSneakyDuck:
New information on various user groups, crowdsourced via https://www.reddit.com/r/freebsd/comments/1sskgr5/

```
Updated:
Portland   new website with contact details, remove outdated email,
           schedule confirmed still correct by u/patmaddox on Reddit

Removed:
Ulm        website live but states group has disbanded
UK         dead link, defunct according to Sam Smith via email
Manchester squatted link, confirmed defunct by Sam Smith via email
Seattle    dead link, confirmed defunct by u/linkslice on Reddit
Jogja      dead email address, no link, no web presence found

```

Re Portland: the new website contains X/Mastodon accounts for two organisers as a means of contact, replacing the previous email address. Pat Maddox suggests it's better to let their website be the source of truth on contact details in case of future changes, rather than list those social media accounts here. The group uses Calagator similarly to how some other groups use meetup.com; adding it here provides some redundancy to check if the group is still active even if their site goes dead.

Re Jogja: this is the only removed group I wasn't able to get direct confirmation for. But the sole information provided when this entry was added 20+ years ago was a university student email address, which has likely been dead for decades. I could find no trace of a web presence.

The list has many other dead links that suggest groups have disbanded, though - as Portland proves - information here can become stale and sometimes contact details have just changed! I intend to ask around for further confirmations on other forums I'm active in. Several defunct groups also appear on the equivalent NetBSD and OpenBSD lists, whose maintainers I intend to notify.

My search may throw up user groups not listed here. I was informed BSD Sul in Brazil, previously removed from this list, has reactivated and participates in activities like BSD Day. Unfortunately their new site has no contact information and is effectively a personal BSD blog: https://bsdsul.com.br/?action=view&url=voltamos-depois-de-oito-anos

My plan for new discoveries is to suggest they file a Bugzilla PR with website plus contact details, as per the instructions at https://www.freebsd.org/usergroups/ ... but I'm aware instructions can be outdated. Are Bugzilla PRs preferable these days?

### DIFF
--- a/website/content/en/usergroups/_index.adoc
+++ b/website/content/en/usergroups/_index.adoc
@@ -79,13 +79,8 @@ link:http://berklix.org/bim/[Berkeley in Munich (BIM)]::
 The Berkeley in Munich (BIM) caters for users of BSD based systems in Oberbayern.
 Located in Germany, Munich.
 
-link:http://www.u-bus.de[Ulmer BSD User Stammtisch]::
-The U-BUS meets every last Thursday of the month.
-For more information about meetings and related issues, please check the web page (link:http://www.u-bus.de[u-bus]).
-Located in Germany, Ulm.
-
 link:http://www.bluefrogs.de/[Bluefrogs e.V.]::
-Unix and Linux User Group in Bergisch Gladbach (near Cologne), German.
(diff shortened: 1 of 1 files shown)


sergio-carlavilla:
Done!

PR #663, opened 2026-04-28
usergroups: Update venues for Hamburg and Portland

BigSneakyDuck:
My crowdsourced request for updates on usergroups has found two more changes.

Hamburg changed venue after 30 years. Include link to their meeting page and update venue to match. Reported by cracauer@ on https://forums.freebsd.org/threads/102512/#post-757478

Portland website has updated their meeting link. Calagator search now quotes "BSD Pizza Night". This removes false positives.


concussious (approved):
Thank you, this is beautiful. The usergroups page needs a lot of love, but it's very involved work and I don't have the bandwidth to go talk to all the usergroups. Please keep sending these patches. The citations make them very easy to verify and merge.

BigSneakyDuck:
> The citations make them very easy to verify and merge.

Thanks - I have a few more in the pipeline. The tricky thing is knowing whether a group has disbanded, or a dead link just means they've switched to alternative ways of organising. In some cases I might cite personal correspondence (generally by email) which obviously you can't verify directly, but that would usually be to confirm a group is inactive so it's still possible to check that their web presence has gone silent.

PR #672, opened 2026-05-19
hardware: update pSeries entries

mchoo7:
14.4 and 15.1 support little-endian pSeries.
15.0 has boot panic due to bug in SLOF, but it should be reclassified as QEMU pSeries instead of IBM.


mchoo7:
replaced by https://reviews.freebsd.org/D57102

PR #680, opened 2026-06-04
better us of ls in example

janstary:
To show the sticky bit on the /tmp directory,
just ls the /tmp directory, as opposed to
listing everything and greping for /tmp


concussious:
Hey Jan! Thanks for taking a look at this. I think `ls -al` is one of the most common use-cases that people should be introduced to in basics, whereas I actually had to look `-d` up because I never use it in 20 years, so this is not really the direction I'd like to move in.

concussious:
Since it's been a few months and nobody else has said anything, I'm going to go ahead and close this. Thanks for your interest, please let us know if you have any other ideas to improve the docs!

PR #703, opened 2026-07-27
handbook: fix lagg example layout and add missing bsdinstall image

branchcat:
- Keep the notes in Handbook section 35.9.1, Example 3, inside the
  example block so that it renders in the intended location.
- Add the missing `bsdinstall-distributions-sets-pkgbase.png` image
  used by Figure 8 of Handbook section 2.5.3.


concussious (approved):

concussious:
Thanks!

PR #704, opened 2026-08-03
Fix HTTP to HTTPS link for Wi-Fi Alliance in advanced-networking.

siddhardh-codemonk:
Update the Wi-Fi Alliance link from HTTP to HTTPS in the advanced-networking chapter.


concussious:
Thanks for sending this in. I do not want to take this because, correct me if I'm wrong, https provides no benefit for this use case:

1. the Wi-Fi alliance site provides no fields for user credentials to be submitted, therefore there is no need to encrypt them because they could not be intercepted.
2. browsers on high performance devices will automatically use https anyway
3. low performance devices may not be able to calculate https encryption

